PR #2, opened 2018-02-21
CO-1019 Implement persons api

tsoliangwu0130:


swagger.yaml
```diff
@@ -39,23 +119,117 @@ parameters:
...
+                      type: string
+                      format: date
+                      description: Effective date of job.
+                    status:
+                      type: string
+                      description: Job staus
+                      enum:
+                        - Active
+                        - Leave without pay with benefits
+                        - Leave without pay without benefits
```
jaredkosanovic: Can you change this to "Leave without pay and benefits"?

swagger.yaml
```diff
@@ -212,8 +212,8 @@ definitions:
                       description: Job staus
                       enum:
                         - Active
-                        - Leave without pay with benefits
-                        - Leave without pay without benefits
+                        - Leave without pay but with benefits
```
jaredkosanovic: This looks good, but don't forget to update these strings in the mapper as well.

src/main/groovy/edu/oregonstate/mist/core/PersonObject.groovy
```diff
@@ -0,0 +1,23 @@
...
+import com.fasterxml.jackson.annotation.JsonIgnore
+
+class PersonObject {
+    @JsonIgnore
+    String osuID
+    String lastName
+    String homePhone
+    String alternatePhone
+    String osuUID
+    Date birthDate
```
jaredkosanovic: For all the dates, when I did some simple initial testing, it looks like the date format happened to adhere to [ISO8601 standards](https://xml2rfc.tools.ietf.org/public/rfc/html/rfc3339.html#anchor14) for a "full-date", which is what we want our API to produce. However, I'm thinking we should explicitly state declare this format instead of assuming that our API will produce it. [Here's a similar method used for the locations-api](https://github.com/osu-mist/locations-api/blob/develop/src/main/groovy/edu/oregonstate/mist/locations/core/DayOpenHours.groovy#L9). I'm thinking we want to change all date fields for the persons API (for all endpoints that have dates) to declare the intended format in the code.

Another thought I had is whether we want to produce dates in UTC, like we do for the locations api and some other APIs. There are a few complications to this:
- In Banner, most date fields are of the `Date` datatype, which has no concept of timezone, so we'd have to assume the timezone of all dates (PST or PDT depending on the date). This would be a configuration parameter. I did a similar thing with the OSU Events API (not the message api) except I was dealing with a different backend system (localist rather than banner/oracle) https://github.com/osu-mist/osu-events-api/blob/develop/src/main/groovy/edu/oregonstate/mist/osuevents/resources/EventsResource.groovy#L45-L50
- I'm not even sure if banner keeps a record of time for fields like birth dates, and start and end dates of jobs. If they don't, or if the times aren't reliable, I don't know if we should convert them to UTC. For example, if someone's job start date is January 1st 2018 at 6:00 pm, the UTC date for this would be January 2nd. 

What are your thoughts @tsoliangwu0130 ?

src/main/groovy/edu/oregonstate/mist/personsapi/mapper/JobsMapper.groovy
```diff
@@ -0,0 +1,34 @@
...
+            'T': 'Terminated'
+        ]
+
+        new JobObject(
+            positionNumber: rs.getString('POSITION_NUMBER'),
+            fte: rs.getFloat('FTE'),
+            description: rs.getString('DESCRIPTION'),
+            beginDate: rs.getDate('BEGIN_DATE'),
+            endDate: rs.getDate('END_DATE'),
+            status: jobStatusDict[rs.getString('STATUS')]
```
jaredkosanovic: What if Banner adds a new status letter? Would this line throw an exception since the map wouldn't have the new status? I'm thinking for safety, we should return the literal status letter if the letter is not in `jobsStatusDict`. We should then have an integration test to make sure this field is more than just one character.

src/main/groovy/edu/oregonstate/mist/personsapi/mapper/PersonMapper.groovy
```diff
@@ -0,0 +1,33 @@
...
+            osuID: rs.getString('OSU_ID'),
+            firstName: rs.getString('FIRST_NAME'),
+            lastName: rs.getString('LAST_NAME'),
+            middleName: rs.getString('MIDDLE_NAME'),
+            birthDate: rs.getDate('BIRTH_DATE'),
+            primaryAffiliation: rs.getString('PRIMARY_AFFILIATION'),
+            email: rs.getString('EMAIL_ADDRESS'),
+            username: rs.getString('USERNAME'),
+            osuUID: rs.getString('OSUUID'),
+            confidential: rs.getBoolean('CONFIDENTIAL'),
```
jaredkosanovic: For all these `getBoolean` calls, be careful because it looks like that method only returns true if the column contains a 1: https://docs.oracle.com/javase/7/docs/api/java/sql/ResultSet.html#getBoolean(java.lang.String)

Since boolean/ind fields in banner are often represented by `Y` or `null`, or `Y` or `N`, it might be better to explicitly check for a certain value instead of using `getBoolean`. 

As an example, here is what I did for the students API: https://github.com/osu-mist/students-api/commit/a765a49a10e5b984fdd769005b6606635ac2952e#diff-b4c2ee6fa8bbe8377a138dd3912be159R14

jaredkosanovic: For integration tests, we'll want to have a test to check that a person we know is confidential in banner is being represented as confidential in our API.

tsoliangwu0130: Good catch, thanks. I mapped the `currentStudent` field by calling `getString` and check if the field is equal to `Y`. However, I didn't change `confidential` and `currentUser` mapping methods since these indicators using `1` and `0` here. Do you think I should also explicitly check these two filed even they are already convertible?

jaredkosanovic: We discussed this in person, but to summarize, I don't think we need to explicitly check unless the column is something other than 1 or null, or 1 or 0, since those values would correctly map to true or false in the getBoolean method.

src/main/groovy/edu/oregonstate/mist/personsapi/PersonsResource.groovy
```diff
@@ -101,14 +103,22 @@ class PersonsResource extends Resource {
     @Produces('image/jpeg')
     @Path('{osuID: [0-9]{9}}/image')
     Response getImageById(@PathParam('osuID') String osuID, @QueryParam('width') Integer width) {
-        def image = personsDAO.getImageById(osuID)
 
-        if (image) {
+        if (personsDAO.getPersonById(osuID)) {
```
jaredkosanovic: This conditional is doing a pretty big query (involves subqueries and joins)  just to get a true or false answer on whether a person exists. I would recommend having a separate `personExists` query (or something similar) to do a really simple check to see if the person exists.

src/main/groovy/edu/oregonstate/mist/personsapi/PersonsResource.groovy
```diff
@@ -80,8 +81,9 @@ class PersonsResource extends Resource {
     @GET
     @Path('{osuID: [0-9]{9}}/jobs')
     Response getJobsById(@PathParam('osuID') String osuID) {
-        def jobs = personsDAO.getJobsById(osuID)
-        if (jobs) {
+
+        if (personsDAO.getPersonById(osuID)) {
```
jaredkosanovic: Same comment as the one for id image endpoint: This should be doing a more simple query to see if someone exists.

src/main/groovy/edu/oregonstate/mist/personsapi/PersonsResource.groovy
```diff
@@ -0,0 +1,127 @@
...
+    @Path('{osuID: [0-9]{9}}/jobs')
+    Response getJobsById(@PathParam('osuID') String osuID) {
+
+        if (personsDAO.getPersonById(osuID)) {
+            def jobs = personsDAO.getJobsById(osuID)
+            def res = new ResultObject(
+                data: new ResourceObject(
+                    id: osuID,
+                    type: 'jobs',
+                    attributes: jobs,
```
jaredkosanovic: The attributes object in geojson must be an object: http://jsonapi.org/format/#document-resource-object-attributes

In the [swagger spec](https://github.com/osu-mist/persons-api/blob/feature/CO-1019-implement-persons-api/swagger.yaml#L196), there is a `jobs` element whose value would be the jobs array:
```
{
  "links": {},
  "data": {
    "id": "foobar",
    "type": "jobs",
    "attributes": {
      "jobs" : [
        {
        "beginDate": "2016-08-22",
        "endDate": null,
        "positionNumber": "CFOOBAR",
        "status": "Active",
        "description": "Analyst Programmer",
        "fte": 1
      } etc. etc.
```

src/main/groovy/edu/oregonstate/mist/personsapi/PersonsResource.groovy
```diff
@@ -0,0 +1,127 @@
...
+                    )
+                }
+            )
+            ok(res).build()
+        }
+    }
+
+    @Timed
+    @GET
+    @Path('{osuID: [0-9]{9}}')
```
jaredkosanovic: I'm hesitant about declaring the exact number of digits the ID should be. One side effect right now is that if I search with a 9 digit id that doesn't exist in the database, I will get a 404 with the application/json header. If I search with an id that's not 9 digits, it won't even invoke the method and the 404 it returns doesn't have the application/json header. That's an issue either way, but I think it should be fixed in a different pr for the skeleton. In the mean time, perhaps we shouldn't validate the length of the id?

src/main/groovy/edu/oregonstate/mist/personsapi/PersonsResource.groovy
```diff
@@ -0,0 +1,127 @@
...
+
+    @Timed
+    @GET
+    Response list(@QueryParam('onid') String onid,
+                  @QueryParam('osuID') String osuID,
+                  @QueryParam('osuUID') String osuUID) {
+
+        def id = [onid, osuID, osuUID].findAll { it }
+        if (id.size() != 1) {
+            badRequest('The number of input parameter(s) is not equal to one.').build()
```
jaredkosanovic: I think we should make this error message more specific. We might be adding more query parameters to this API - and those parameters might be allowed to be included together. We could say something like "onid, osuID, and osuUID cannot not be included together"

src/main/groovy/edu/oregonstate/mist/personsapi/mapper/PersonMapper.groovy
```diff
@@ -0,0 +1,33 @@
...
```
jaredkosanovic: I tried doing a query with a confidential person and it threw a 500 error. I think it's because the confidential indicator uses `Y` or `null`, not 1 or 0.

src/main/groovy/edu/oregonstate/mist/personsapi/PersonsResource.groovy
tsoliangwu0130: Good point.

PR #6, opened 2018-04-04
Feature/co 1065 post jobs endpoint

jaredkosanovic:


swagger.yaml
```diff
@@ -14,6 +14,8 @@ schemes:
 basePath: /v1/persons
```
tsoliangwu0130: According to the [swagger document](https://swagger.io/docs/specification/2-0/api-host-and-base-path/), the basePath should be `/v1`, and the path below should also be changed correspondingly, e.g. `/` -> `/persons`. Also, could you also add `host` field which points to the prod environment? It will be helpful when we wanna create API document by importing the swagger specification directly to SmartDoc.

PR #7, opened 2018-04-09
Implement fuzzy name searching

jaredkosanovic:
This adds the ability to do fuzzy name searches based on first and last name. This also allows searching based on previous OSU ID's or names.

### DIFF
--- a/src/main/groovy/edu/oregonstate/mist/personsapi/PersonsResource.groovy
+++ b/src/main/groovy/edu/oregonstate/mist/personsapi/PersonsResource.groovy
@@ -46,20 +46,20 @@ class PersonsResource extends Resource {
         // Check for a bad request.
         Closure<Integer> getSize = { List<String> list -> list.findAll { it }.size() }
 
-        Integer ids = getSize([onid, osuID, osuUID])
-        Integer names = getSize([firstName, lastName])
+        Integer idCount = getSize([onid, osuID, osuUID])
+        Integer nameCount = getSize([firstName, lastName])
 
-        Boolean validNameRequest = names == 2
+        Boolean validNameRequest = nameCount == 2
 
         String errorMessage
 
         if (osuUID && !osuUID.matches("[0-9]+")) {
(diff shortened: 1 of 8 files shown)


swagger.yaml
```diff
@@ -149,6 +165,24 @@ definitions:
...
+                      example: "931234567"
+                    firstName:
+                      type: string
+                      description: Previous first name
+                    middleName:
+                      type: string
+                      description: Previous middle name
+                    lastName:
+                      type: string
+                      description: Previous last name
```
tsoliangwu0130: I forgot did we discuss this in person or not. Could we also add the modification date for each record?

swagger.yaml
```diff
@@ -35,6 +35,22 @@ paths:
...
+          name: lastName
+          required: false
+          description: Last name of a person, for fuzzy name searching.
+          type: string
+        - in: query
+          name: searchOldVersions
+          required: false
+          default: false
+          description: If true, searching by name or OSU ID will also search old names and OSU ID's associated with the person.
+          type: boolean
```
tsoliangwu0130: Another thought, do you think it will be valuable if we can provide the feature that allows user search by old name/OSU ID?

src/main/groovy/edu/oregonstate/mist/personsapi/PersonsResource.groovy
```diff
@@ -33,57 +33,101 @@ class PersonsResource extends Resource {
...
+        person.previousRecords = personsDAO.getPreviousRecords(person.internalID)
+
+        new ResourceObject(
+                id: person.osuID,
+                type: 'person',
+                attributes: person,
+                links: ['self': personUriBuilder.personUri(person.osuID)]
+        )
+    }
+
```
tsoliangwu0130: Great work 👍 nice and neat!

jaredkosanovic: 😄

PR #11, opened 2018-05-09
Feature/co 1082 post endpoint

jaredkosanovic:


tsoliangwu0130:
* When sending POST request without any data, I got 500 and uncaught exceptions
* When creating jobs with text/file type, I got 415 Unsupported Media Type but not followed JSON API format
* When I sent the date data with this wrong format `2018-05-01-xxxxxxx`, the validation didn't catch the error and send me a 202
* When I entered the "status" with a number or other meaningless strings like "Hello world", I also got a 202
* `hoursPerPay` and `hourlyRate` accepted negative numbers
* `assignmentSalary` not necessarily equal to `hoursPerPay` times `hourlyRate`
* `timesheetOrganizationCode` accepted random strings as well
* When sending some bad requests, like creating multi jobs in one POST request or sending a string in "annualSalary" field, I did get a 400 but the error message is "Could not parse job object. Make sure dates are in ISO8601 format: yyyy-MM-dd"

PR #15, opened 2018-07-09
add name and old osu id tests

jaredkosanovic:
CO-1078

### DIFF
--- a/src/test/integration/integrationtests.py
+++ b/src/test/integration/integrationtests.py
@@ -49,8 +49,8 @@ class TestStringMethods(unittest.TestCase):
 
         # identifiers and return ids should matched
         person_data = osu_id_res.json()['data'][0]
-        ids_list = [(person_data['id'],
-                     osu_id), (person_data['attributes']['username'], onid),
+        ids_list = [(person_data['id'], osu_id),
+                    (person_data['attributes']['username'], onid),
                     (person_data['attributes']['osuUID'], osuuid)]
         for res_id, param_id in ids_list:
             self.assertEqual(res_id, param_id)
@@ -130,8 +130,8 @@ class TestStringMethods(unittest.TestCase):
         no_job_res = utils.get_jobs_by_osu_id(no_job_osu_id)
 
         # expect 200 if osuID is valid
-        for res, res_osu_id in [(jobs_res, jobs_osu_id), (no_job_res,
-                                                          no_job_osu_id)]:
+        for res, res_osu_id in [(jobs_res, jobs_osu_id),
+                                (no_job_res, no_job_osu_id)]:
             self.assertEqual(res.status_code, 200)
 
         # test person with jobs
--- a/src/test/integration/integrationtests.py
+++ b/src/test/integration/integrationtests.py
@@ -216,10 +216,8 @@ class TestStringMethods(unittest.TestCase):
 
     def test_name_search(self):
         person_res = utils.get_person_by_ids({
-            'firstName':
-            regular_name_person['first_name'],
-            'lastName':
-            regular_name_person['last_name']
+            'firstName': regular_name_person['first_name'],
+            'lastName': regular_name_person['last_name']
         })
 
         self.assertEqual(person_res.status_code, 200)
@@ -228,10 +226,8 @@ class TestStringMethods(unittest.TestCase):
 
     def test_fuzzy_name_search(self):
         person_res = utils.get_person_by_ids({
-            'firstName':
-            fuzzy_name_person['first_name'],
-            'lastName':
-            fuzzy_name_person['last_name']
+            'firstName': fuzzy_name_person['first_name'],
+            'lastName': fuzzy_name_person['last_name']
         })
 
         self.assertEqual(self.length_of_response(person_res), 1)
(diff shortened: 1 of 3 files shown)


src/test/integration/integrationtests.py
```diff
@@ -167,18 +201,89 @@ def test_self_link(self):
...
 
             self.assertEqual(job['links']['self'], expected_url)
 
+    def test_name_search(self):
+        person_res = utils.get_person_by_ids({
+            'firstName':
+            regular_name_person['first_name'],
+            'lastName':
+            regular_name_person['last_name']
+        })
```
tsoliangwu0130: I think these lines aren't long enough to break to new lines. Could you change the code like you did in [line 261](https://github.com/osu-mist/persons-api/blob/89462c618541f3706117dd844adb532edb3d23e9/src/test/integration/integrationtests.py#L264-L267)?

```python
person_res = utils.get_person_by_ids({
    'firstName': regular_name_person['first_name'],
    'lastName': regular_name_person['last_name']
})
```

src/test/integration/integrationtests.py
```diff
@@ -167,18 +201,89 @@ def test_self_link(self):
...
+        self.assertEqual(self.length_of_response(person_res), 1)
+
+    def test_fuzzy_name_search(self):
+        person_res = utils.get_person_by_ids({
+            'firstName':
+            fuzzy_name_person['first_name'],
+            'lastName':
+            fuzzy_name_person['last_name']
+        })
+
```
tsoliangwu0130: Same here.

src/test/integration/integrationtests.py
```diff
@@ -167,18 +201,89 @@ def test_self_link(self):
...
+    def test_name_alias(self):
+        for person_name in alias_persons:
+            person_res = utils.get_person_by_ids({
+                'firstName':
+                person_name['first_name_alias'],
+                'lastName':
+                person_name['last_name']
+            })
+
```
tsoliangwu0130: Same here.

src/test/integration/integrationtests.py
```diff
@@ -33,31 +49,33 @@ def test_person_by_ids(self):
...
 
     def test_phones(self):
         phones_res = utils.get_person_by_osu_id(phones_osu_id)
 
         home_phone = phones_res.json()['data']['attributes']['homePhone']
-        alternate_phone = phones_res.json()['data']['attributes']['alternatePhone']
+        alternate_phone = phones_res.json()['data']['attributes'][
+            'alternatePhone']
         primary_phone = phones_res.json()['data']['attributes']['primaryPhone']
         mobile_phone = phones_res.json()['data']['attributes']['mobilePhone']
```
tsoliangwu0130: For this block, could you store `phones_res.json()` as a variable (or even `phones_res.json()['data']['attributes']`) so that we won't need to call `.json()` function every time and also could make the code neater?

jaredkosanovic: Sure, I was using google's yapf to format the code which is why these lines are changed.

src/test/integration/integrationtests.py
```diff
@@ -33,31 +49,33 @@ def test_person_by_ids(self):
...
         # identifiers and return ids should matched
         person_data = osu_id_res.json()['data'][0]
-        ids_list = [
-            (person_data['id'], osu_id),
-            (person_data['attributes']['username'], onid),
-            (person_data['attributes']['osuUID'], osuuid)
-        ]
+        ids_list = [(person_data['id'],
+                     osu_id), (person_data['attributes']['username'], onid),
+                    (person_data['attributes']['osuUID'], osuuid)]
```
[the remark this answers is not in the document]
jaredkosanovic: Regarding all the formatting comments, I ran the code through google's [yapf](https://github.com/google/yapf) with the default settings, which formats to pep8. I confirmed it conforms to pep8 by running this tool: https://github.com/PyCQA/pycodestyle

I want to get out of the business of us manually formatting our code. For example, go has [gofmt](https://golang.org/cmd/gofmt/) to help with this, and I figured yapf could do the similar task for python.

I'm going to leave all these formatting comments as-is right now since the code conforms to pep8. Perhaps in the future we can talk about how we want to agree to use something like yapf so the format is something we can agree on.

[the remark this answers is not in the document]
tsoliangwu0130: I'm not opposed to using yapf or similar tools to format the code, but I found that sometimes these tools format the code in a kind of weird way, for example, breaking the lines even it is not really necessary. Similarly, although we also use codenarc to auto-checked the code style for our Groovy codes, we still always be very careful of the coding style when we reviewing the code. The reason behind that is because I think the readability is also very important. In my opinion, it might directly influence the difficulty of understanding.
However, I also understand that coding style could be various depends on persons. That is probably why I think we should discuss the styling stuff whenever doing the code review. We don't really want to wait until the code becomes too large. What do you think?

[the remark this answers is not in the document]
jaredkosanovic: Yeah, that's fair. I'll make a few changes. I think we can agree that ideally, we wouldn't have to manually format code at all. I'm starting to see why Andrew liked go so much. Among other reasons, gofmt seems to be one of the best formatters out there. I agree yapf did do some questionable things.

PR #16, opened 2018-07-20
Feature/co 1179 add current employee

jaredkosanovic:


src/test/integration/integrationtests.py
```diff
@@ -274,6 +274,29 @@ def validate_old_parameters(self, parameters, old_param):
...
+    def test_current_employee(self):
+        employee = utils.get_person_by_osu_id(jobs_person).json()
+        assertTrue(self.get_current_employee_field(employee))
+
+        not_employee = utils.get_person_by_osu_id(no_job_osu_id).json()
+        assertFalse(self.get_current_employee_field(not_employee))
+
+    @staticmethod
+    def get_current_employee_field(person):
+        return person['data']['attributes']['currentEmployee']
```
ianbrown9475: Should you validate that the status codes of `employee` and `not_employee` are 200 before getting their `currentEmployee` status or are you assuming valid_ids for these from the configuration file? I would actually like to know myself as far as how much you can trust the values given in the configuration file. I think it would make sense to validate the status code though so that someone testing would be able to more easily see what went wrong instead of getting a key error exception. @jaredkosanovic what do you think about this?

src/test/integration/utils.py
```diff
@@ -58,3 +59,14 @@ def get_jobs_by_osu_id(osu_id):
 def get_image_by_osu_id(osu_id, params=None):
     global api_url, headers
     return requests.get(api_url + osu_id + '/image', headers=headers, params=params)
```
ianbrown9475: Not your fault but there's a pep 8 error here: `utils.py:61:80: E501 line too long (84 > 79 characters)`

src/test/integration/integrationtests.py
jaredkosanovic: Good point. I'll add the check for a `200`.

src/test/integration/integrationtests.py
```diff
@@ -274,6 +274,29 @@ def validate_old_parameters(self, parameters, old_param):
...
+
+        for meal_plan in meal_plan_response.json()['data']:
+            single_meal_plan_response = utils.get_meal_plan_by_id(
+                    meal_plan_person, meal_plan['id'])
+            self.assertEqual(single_meal_plan_response.status_code, 200)
+            self.assertEqual(
+                    meal_plan, single_meal_plan_response.json()['data'])
+
+    def test_current_employee(self):
+        employee = utils.get_person_by_osu_id(jobs_person).json()
```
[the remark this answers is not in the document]
jaredkosanovic: Yes :)

PR #20, opened 2018-08-29
CO-1112: Post jobs integration tests

ianbrown9475:


swagger.yaml
```diff
@@ -476,7 +476,7 @@ definitions:
                   type: string
                 activityCode:
                   type: string
-                distributionPercentage:
+                distributionPercent:
```
jaredkosanovic: Good catch.

src/test/integration/README.md
```diff
@@ -1,6 +1,6 @@
 # Persons API Integration Tests
 
-The integration tests for Persons API.
+The integration tests for Persons API written using Python 2.
```
jaredkosanovic: But our dockerfile uses python 3? https://github.com/osu-mist/persons-api/blob/6272918587325953d38ec4b82809cf7287e1ae70/src/test/integration/Dockerfile#L1

src/test/integration/integrationtests.py
```diff
@@ -144,6 +147,136 @@ def test_jobs_by_osu_id(self):
         self.assertEqual(
             utils.get_person_by_osu_id(not_valid_osu_id).status_code, 404)
 
+    def test_post_job_by_osu_id(self):
```
jaredkosanovic: This method is pretty long. Could you split this up in to several different tests?

src/test/integration/integrationtests.py
```diff
@@ -144,6 +147,136 @@ def test_jobs_by_osu_id(self):
...
+        self.assertEqual(len(no_body.json()), 1)
+        self.assertIn("No job object provided",
+                      no_body.json()[0]["developerMessage"])
+
+        # dates are not in ISO8601 format
+        self.validate_bad_job_post(test_job, "effectiveDate", "badDate",
+                                   "Make sure dates are in ISO8601 format")
+
+        # required field is missing
+        required_fields = ["positionNumber", "hoursPerPay", "assignmentSalary",
```
jaredkosanovic: These should be the required fields https://github.com/osu-mist/persons-api/blob/076f08aaecc8491d578834aa0a6e469023331807/src/main/groovy/edu/oregonstate/mist/personsapi/PersonsResource.groovy#L251-L255

jaredkosanovic:
I ran these tests and the test_post_job test failed
```
======================================================================
FAIL: test_post_job_by_osu_id (__main__.TestStringMethods)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "integrationtests.py", line 164, in test_post_job_by_osu_id
    no_body.json()[0]["developerMessage"])
AssertionError: 'No job object provided' not found in 'Could not parse job object. Make sure dates are in ISO8601 format: yyyy-MM-dd'

----------------------------------------------------------------------
Ran 17 tests in 19.750s
```

I think it will be easier to test these tests if the test_post_job test is split up into separate tests (_how many times can I say test in one sentence_). I will re-run the tests during the next code review.

src/test/integration/README.md
ianbrown9475: Oh you're right. For some reason I assumed it was Python 2. I'll change this.

swagger.yaml
```diff
@@ -438,7 +439,7 @@ definitions:
           annualSalary:
             type: number
             format: float
-            example: 20,050.35
+            example: 20050.35
```
jaredkosanovic: good catch!

src/test/integration/integrationtests.py
```diff
@@ -144,6 +147,125 @@ def test_jobs_by_osu_id(self):
...
+
+        # invalid fields
+        invalid_fields = [
+            {"supervisorOsuID": "Supervisor OSU ID does not exist"},
+            {"positionNumber": "not a valid position number for the given "
+                               "begin date"},
+            {"locationID": "not a valid location ID"},
+            {"timesheetOrganizationCode": "not a valid organization code"}
+        ]
+        for field, message in invalid_fields:
```
jaredkosanovic: This line returned this error: 
```
======================================================================
ERROR: test_bad_post_job (__main__.TestStringMethods)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "integrationtests.py", line 185, in test_bad_post_job
    for field, message in invalid_fields:
ValueError: not enough values to unpack (expected 2, got 1)

----------------------------------------------------------------------
```

Maybe `invalid_fields` should just be one dict instead of an array of dicts?

ianbrown9475: Ah right, thanks for the catch

src/test/integration/integrationtests.py
```diff
@@ -171,17 +171,17 @@ def test_bad_post_job(self):
         positive_fields = ["hourlyRate", "hoursPerPay", "assignmentSalary",
                            "annualSalary", "paysPerYear"]
         for field in positive_fields:
-            self.validate_bad_job_post(test_job, field, "-1",
+            self.validate_bad_job_post(test_job, field, -1,
```
ianbrown9475: I did some testing to confirm that the api doesn't have a problem with these values being strings but I changed it to a number to follow the documentation

src/test/integration/integrationtests.py
```diff
@@ -171,17 +171,17 @@ def test_bad_post_job(self):
...
-            {"timesheetOrganizationCode": "not a valid organization code"}
-        ]
+        invalid_fields = {
+            "supervisorOsuID": "Supervisor OSU ID does not exist",
+            "positionNumber": "not a valid position number for the given "
+                               "begin date",
+            "locationID": "not a valid location ID",
+            "timesheetOrganizationCode": "not a valid organization code",
+        }
         for field, message in invalid_fields:
```
jaredkosanovic: Ah, this should be `invalid_fields.items()`

PR #42, opened 2018-11-19
add i9 fields

jaredkosanovic:


swagger.yaml
```diff
@@ -622,6 +622,17 @@ definitions:
         description: The date that the employee accepted or denied to have their salary information shared.
       retirementCode:
         type: string
+      i9FormCode:
```
jaredkosanovic: this field will rarely be used and there isn't a validation table for the various codes :(. We can always add a `i9FormDescription` field in the future if we clients want a more human readable value, but for now it would be nice to not have the added maintenance.

swagger.yaml
```diff
@@ -622,6 +622,17 @@ definitions:
...
+        type: string
+        description: Code to indicate the status of the employee's I9 form.
+      i9Date:
+        type: string
+        format: date
+        description: The date that the I9 form was received.
+      i9ExpirationDate:
+        type: string
+        format: date
+        description: The date the I9 form expires.
```
tsoliangwu0130: Could you give examples of the dates? It's mainly for users to know what format of date will get by just reading the swagger/openapi specification.

swagger.yaml
tsoliangwu0130: For `i9FromCode`, is it possible to get all possible values? Also, is there isn't a validation table, do you know which table we can use to get the descriptions?

jaredkosanovic: @tsoliangwu0130 There isn't a validation table, so there isn't a table we can get the descriptions from :(

swagger.yaml
jaredkosanovic: I'd rather leave that up to the client using the specification. `format: date` specifically refers to this format: 2018-12-31 https://tools.ietf.org/html/rfc3339#section-5.6

For example, the swagger editor will auto populate an example date in the example json based on the date format (`date` or `date-time`)

PR #55, opened 2019-06-26
CO-1465 add supervisor org data

tsoliangwu0130:


tsoliangwu0130:
Actually the ticket number should be CO-1478.

PR #58, opened 2019-08-14
CO-1561 add ssn status

tsoliangwu0130:


src/main/groovy/edu/oregonstate/mist/personsapi/mapper/PersonMapper.groovy
```diff
@@ -14,6 +14,17 @@ import java.sql.SQLException
 public class PersonMapper implements ResultSetMapper<PersonObject> {
     PhoneFormatter phoneFormatter = new PhoneFormatter()
     public PersonObject map(int i, ResultSet rs, StatementContext sc) throws SQLException {
+        String rawSSN = rs.getString('SSN')
+        String ssnStatus = 'invalid'
+
+        if (rawSSN && rawSSN.startsWith('VAULT')) {
+            ssnStatus = 'valut'
```
ianbrown9475: Typo `valut` -> `vault`

src/main/groovy/edu/oregonstate/mist/personsapi/mapper/PersonMapper.groovy
```diff
@@ -14,6 +14,17 @@ import java.sql.SQLException
...
+        } else if (rawSSN == 'REDACTED' || rawSSN == 'ARCHIVED') {
+            ssnStatus = rawSSN.toLowerCase()
+        } else if (rawSSN && rawSSN.matches(/^(?!0{9})\d{9}$/)) {
+            ssnStatus = 'valid'
+        }
```
ianbrown9475: Would it make sense to move this logic to a private method of this class, similar to `formatPhoneNumber()`?

ianbrown9475: Does this logic handle `rawSSN` being `NULL`?

ianbrown9475: What should happen if `rawSSN` doesn't match any of the cases here? Should we throw an exception or something?

ianbrown9475: This might be a little overly paranoid, but would we do something like set `rawSSN` to `null` after the logic for getting `ssnStatus` has completed to make sure this variable's value is never used anywhere else? I guess this wouldn't be necessary if we move the logic to its own private function, since it would be freed from memory when the method scope is exited, right?

src/test/integration/requirements.txt
```diff
@@ -4,4 +4,4 @@ idna==2.6
 phonenumbers==8.9.2
 Pillow==5.0.0
```
ianbrown9475: This version isn't supported by Python 3.7: https://pillow.readthedocs.io/en/latest/installation.html#notes The installation will fail with pip. We should probably either update the Docker python version or update this package version

src/main/groovy/edu/oregonstate/mist/personsapi/mapper/PersonMapper.groovy
tsoliangwu0130: yeah, you're right, will do.

tsoliangwu0130: I think so. If `rawSSN` is null, then the status will be `invalid`.

tsoliangwu0130: No, then the `ssnStatus` should be `invalid`.

src/test/integration/requirements.txt
tsoliangwu0130: Good catch, the python 3 docker image in jenkins vm is 3.6 so I didn't run into this issue while testing. Will fix.

src/main/groovy/edu/oregonstate/mist/personsapi/mapper/PersonMapper.groovy
tsoliangwu0130: That's a really good point, I'll think about it more.

src/main/groovy/edu/oregonstate/mist/personsapi/mapper/PersonMapper.groovy
```diff
@@ -14,6 +14,17 @@ import java.sql.SQLException
 public class PersonMapper implements ResultSetMapper<PersonObject> {
     PhoneFormatter phoneFormatter = new PhoneFormatter()
     public PersonObject map(int i, ResultSet rs, StatementContext sc) throws SQLException {
+        String rawSSN = rs.getString('SSN')
+        String ssnStatus = 'invalid'
+
+        if (rawSSN && rawSSN.startsWith('VAULT')) {
+            ssnStatus = 'valut'
+        } else if (rawSSN == 'REDACTED' || rawSSN == 'ARCHIVED') {
```
fernamig: In his email, Reed said "is “REDACTED”, “ARCHIVED”, or starts with “VAULT”, then the SSN is in the vault" so I think this should actually be vault also.

tsoliangwu0130: Right, I originally set them all as `vault` too, but thought about it more, since `REDACTED` and `ARCHIVED` were more specific and clearer statuses than just 'in vault', I eventually decided to have different statuses for them. @fernmig what do you think?

src/main/groovy/edu/oregonstate/mist/personsapi/mapper/PersonMapper.groovy
tsoliangwu0130: @ianbrown9475 I found this article and it shows the best practice about how to store sensitive data into variables: https://dev.to/awwsmm/how-to-encrypt-a-password-in-java-42dh

However, think about it more, I think since we only need to know the SSN staus, I should prevent to get the rawSSN in the first place. Therefore, I move the calculation logic to SQL query, so that we'll always get the ssn status for persons.

src/main/groovy/edu/oregonstate/mist/personsapi/mapper/PersonMapper.groovy
tsoliangwu0130: Reed mentioned `REDACTED` are `ARCHIVED` are all in the vault now, based on our discussion, I'm going to hide these detail codes for now unless there is another need from our client in the future.

src/main/groovy/edu/oregonstate/mist/personsapi/mapper/PersonMapper.groovy
ianbrown9475: This sounds like the best solution

src/main/groovy/edu/oregonstate/mist/personsapi/mapper/PersonMapper.groovy
fernamig: Sounds good. With the changes to the query means this code for ssnStatus can go away and all you need is "ssnStatus = rawSSN.toLowerCase()".

swagger.yaml
```diff
@@ -437,6 +437,15 @@ definitions:
...
                 description: E.164 formatted phone number. Capable of SMS communication.
+              ssnStatus:
+                type: string
+                description: SSN status
+                enum:
+                  - valid
+                  - invalid
+                  - vault
+                  - archived
+                  - redacted
```
fernamig: Remember to update this enum to be only valid, invalid, vault.

tsoliangwu0130: @fernmig sure, you need to refresh the page to view the latest change.

PR #62, opened 2019-09-23
CO-1582 update ssn

tsoliangwu0130:


src/main/groovy/edu/oregonstate/mist/personsapi/PersonsResource.groovy
```diff
@@ -714,6 +714,7 @@ class PersonsResource extends Resource {
                        @Valid ResultObject resultObject) {
 
         String pidm = personsDAO.personExist(osuID)
+        Boolean hasSPBPERS = personsDAO.hasSPBPERS(pidm)
```
fernamig: hasSPBPERS is not used until the try block, so the query could be executed without needing the data. Move initialization of hasSPBPERS down to right before the try block or into the try block.

PR #68, opened 2019-10-04
CO-1605 Fix misleading error message

tsoliangwu0130:


src/main/groovy/edu/oregonstate/mist/personsapi/PersonsResource.groovy
```diff
@@ -466,8 +466,10 @@ class PersonsResource extends Resource {
         try {
             job = JobObject.fromResultObject(resultObject)
         } catch (PersonObjectException e) {
-            addBadRequest("Could not parse job object. " +
-                    "Make sure dates are in ISO8601 format: yyyy-MM-dd")
+            addBadRequest(
+                "Could not parse job object. Please make sure all required fields are included and "
+                + "in the corret format."
```
Ruefa: I think the `+` should follow the end of the first line to be consistent with the rest of the code.

tsoliangwu0130: Sounds good, fixed.

PR #69, opened 2019-10-16
CO-1606 create update person address

tsoliangwu0130:


swagger.yaml
```diff
@@ -867,69 +897,117 @@ definitions:
...
+      data:
+        $ref: "#/definitions/AddressResourceObject"
+  AddressResourceObject:
+    properties:
+      id:
+        type: string
+        description: Globally unique ID for the address.
+      type:
+        $ref: "#/definitions/AddressType"
+      attributes:
```
Ruefa: I see a lot of the attributes here are also in `newAddressResourceObject`. I think it would be a good idea to define these separately and then `$ref` them.

tsoliangwu0130: Good point. Fixed.

tsoliangwu0130:
The conflict is due to the earn code change for [CO-1619](https://jira.sig.oregonstate.edu/browse/CO-1619). I can fix it once this PR got approved.

src/main/groovy/edu/oregonstate/mist/personsapi/PersonsResource.groovy
```diff
@@ -746,6 +749,122 @@ class PersonsResource extends Resource {
...
+                    id: addresses[0].id,
+                    type: "addresses",
+                    attributes: addresses[0]
+                )
+            )).build()
+        } catch (UnableToExecuteStatementException e) {
+            internalServerError("Unable to execute SQL query.").build()
+        } catch (Exception e) {
+            internalServerError(
+                "Internal Server Error, please contact API support team for futher assistance."
```
fernamig: futher => further

src/main/groovy/edu/oregonstate/mist/personsapi/PersonsResource.groovy
```diff
@@ -746,6 +749,122 @@ class PersonsResource extends Resource {
...
+
+        String addressType = resultObject.data['attributes']['addressType']
+        AddressRecordObject addressRecord = bannerPersonsReadDAO.hasSameAddressType(
+            pidm, addressType
+        )
+
+        try {
+            if (addressRecord?.rowID) {
+                logger.info("Address with the same type exist. Deactivate the current one.")
+                bannerPersonsWriteDAO.deactivateAddress(pidm, addressRecord)
```
fernamig: We could deactivate an address and then fail to create a new address. We should reactive the address on a failure.

tsoliangwu0130: @fernmig that's a good point, I thought about it too. But I think the issue here is to deactivate and reactivate is actually calling the same SP. If deactivating running into an issue, it's very possible that reactivating records could run into issue too. And if that's the case, it can cause bad data in the DB. That's why I think if we run into the issue, we need to look at it manually instead of having API try to fix it.

PR #71, opened 2019-10-18
CO-1616 create get phones endpoint

Ruefa:


swagger.yaml
```diff
@@ -959,6 +992,53 @@ definitions:
...
+        items:
+          properties:
+            id:
+              type: string
+              description: Globally unique Id for the phone numbers
+            type:
+              type: string
+              example: phone
+            attributes:
+              properties:
```
tsoliangwu0130: Descriptions of some of these attributes are missing. Can you add descriptions to all of them?

swagger.yaml
```diff
@@ -959,6 +992,53 @@ definitions:
...
+                areaCode:
+                  type: string
+                  example: 541
```
tsoliangwu0130: If a string can be parsed as a number or other types, it should be quoted.

Ref: https://github.com/osu-mist/style-guides/blob/master/OpenAPI.md#quoting-conventions

swagger.yaml
```diff
@@ -959,6 +992,53 @@ definitions:
...
+      data:
+        type: array
+        items:
+          properties:
+            id:
+              type: string
+              description: Globally unique Id for the phone numbers
+            type:
+              type: string
+              example: phone
```
tsoliangwu0130: Should be plural.

swagger.yaml
```diff
@@ -959,6 +992,53 @@ definitions:
...
+                  example: 4443333
+                fullPhoneNumber:
+                  type: string
+                  description: Combination of area code and phone number
+                  example: 5414443333
+                phoneExtension:
+                  type: string
+                  example: 20
+                primaryIndicator:
+                  type: string
```
tsoliangwu0130: This should be a boolean.

swagger.yaml
```diff
@@ -403,6 +413,29 @@ parameters:
...
+    in: query
+    description: |
+      Filter by phone type. Common phone types and their descriptions:
+      - "CM" - Current Mailing
+      - "PA" - Student Alternate Contact
+      - "DM" - Department (US Mail)
+      - "RK" - Financial Aid Address
+      - "RH" - Residence Hall
+    required: false
+    type: string
```
tsoliangwu0130: And it should be enum. I also noticed I forgot to add it too, can you fix it as well? https://github.com/osu-mist/persons-api/blob/ea4efc6976213d31ba59598cf3661db2e853413b/swagger.yaml#L913-L916

swagger.yaml
```diff
@@ -403,6 +413,29 @@ parameters:
...
+    name: addressType
+    in: query
+    description: |
+      Filter by phone type. Common phone types and their descriptions:
+      - "CM" - Current Mailing
+      - "PA" - Student Alternate Contact
+      - "DM" - Department (US Mail)
+      - "RK" - Financial Aid Address
+      - "RH" - Residence Hall
+    required: false
```
tsoliangwu0130: Although it's not required for calling the SP, I think we probably don't want out client to insert a phone without `addressType`. What do you think?

swagger.yaml
```diff
@@ -403,6 +413,29 @@ parameters:
...
+  phoneType:
+    name: phoneType
+    in: query
+    description: |
+      Filter by phone type. Common phone types and their descriptions:
+      - "CM" - Current
+      - "PA" - Student Alternate Contact
+      - "RK" - Financial Aid Address
+      - "MP" - Mobile Phone
+    required: false
```
tsoliangwu0130: It should be required as well.

swagger.yaml
```diff
@@ -403,6 +413,29 @@ parameters:
...
+  addressType:
+    name: addressType
+    in: query
+    description: |
+      Filter by phone type. Common phone types and their descriptions:
+      - "CM" - Current Mailing
+      - "PA" - Student Alternate Contact
+      - "DM" - Department (US Mail)
+      - "RK" - Financial Aid Address
+      - "RH" - Residence Hall
```
tsoliangwu0130: These enum lists can be reused for the result object.

swagger.yaml
```diff
@@ -959,6 +992,53 @@ definitions:
...
+                  type: string
+                  example: Current
+                addressType:
+                  type: string
+                  example: CM
+                  enum: [RN, MG, VO, RE, RB, VP, VF, RK, CM, EF, RH, EO, PA, MP]
+                addressTypeDescription:
+                  type: string
+                  example: Current Mailing
+                activityDate:
```
tsoliangwu0130: Can you rename this as `lastModified` to make it consistent with address endpoint?

src/main/groovy/edu/oregonstate/mist/personsapi/core/PhoneObject.groovy
```diff
@@ -0,0 +1,28 @@
+package edu.oregonstate.mist.personsapi.core
+import edu.oregonstate.mist.api.jsonapi.ResultObject
```
tsoliangwu0130: Can you add a new line after `package`?

src/main/groovy/edu/oregonstate/mist/personsapi/mapper/PhoneMapper.groovy
```diff
@@ -0,0 +1,26 @@
...
+
+public class PhoneMapper implements ResultSetMapper<PhoneObject> {
+  public PhoneObject map(int i, ResultSet rs, StatementContext sc) throws SQLException {
+    new PhoneObject(
+      id: rs.getString('ID'),
+      areaCode: rs.getString('AREA_CODE'),
+      phoneNumber: rs.getString('PHONE_NUMBER'),
+      fullPhoneNumber: rs.getString('AREA_CODE') + rs.getString('PHONE_NUMBER'),
+      phoneExtension: rs.getString('PHONE_EXTENSION'),
+      primaryIndicator: rs.getString('PRIMARY_IND'),
```
tsoliangwu0130: This should be boolean.

src/main/groovy/edu/oregonstate/mist/personsapi/PersonsResource.groovy
```diff
@@ -794,4 +795,29 @@ class PersonsResource extends Resource {
...
     }
+
+    @Timed
+    @GET
+    @Path('{osuID: [0-9]+}/phones')
+    Response getPhones(@PathParam('osuID') String osuID,
+                       @QueryParam('addressType') String addressType,
+                       @QueryParam('phoneType') String phoneType) {
+        if (bannerPersonsReadDAO.personExist(osuID)) {
+            List<PhoneObject> phones = bannerPersonsReadDAO.getPhones(osuID, addressType, phoneType)
```
tsoliangwu0130: We also wanna validate if `addressType` and `phoneType` are valid.

swagger.yaml
```diff
@@ -371,6 +361,26 @@ paths:
...
+        - $ref: '#/parameters/addressType'
+        - $ref: '#/parameters/phoneType'
+        - $ref: '#/parameters/osuID'
+        - $ref: '#/parameters/authorization'
+      responses:
+        200:
+          description: Successful response
+          schema:
+            $ref: '#/definitions/PhonesResultObject'
+        404:
```
tsoliangwu0130: I think we should return a 400 if `phoneType` or `addressType` are not valid.

swagger.yaml
Ruefa: True but these are for the get endpoints. I could be wrong but I don't think these parameters are referenced in the post body definitions.

src/main/groovy/edu/oregonstate/mist/personsapi/PersonsResource.groovy
Ruefa: This validation is not done in `getAddresses`. Would you like me to add that there as well?

tsoliangwu0130: Oh I didn't notice, yeah, feel free to add it if you don't mind, thanks 👍

swagger.yaml
tsoliangwu0130: @Ruefa oh yeah you are right. My brain doesn't work well today 😩 Feel free to ignore my comment here.

swagger.yaml
tsoliangwu0130: Please ignore this as well.

src/main/groovy/edu/oregonstate/mist/personsapi/core/PhoneObject.groovy
```diff
@@ -17,12 +18,12 @@ class PhoneObject {
   String phoneNumber
   String fullPhoneNumber
   String phoneExtension
-  String primaryIndicator
+  boolean primaryIndicator
```
tsoliangwu0130: Should be `Boolean`?

PR #78, opened 2019-11-12
Hotfix CO-1624 fix top level link

Ruefa:


src/main/groovy/edu/oregonstate/mist/personsapi/PersonUriBuilder.groovy
```diff
@@ -48,4 +49,19 @@ class PersonUriBuilder {
                 .queryParam('phoneType', phoneType)
                 .build(osuID)
     }
+
+    URI topLevelUri(UriInfo uri) {
+        UriBuilder builder = UriBuilder.fromUri(this.endpointUri)
+
+        if(uri) {
```
tsoliangwu0130: A space needed between `if` and `(`.

PR #83, opened 2020-02-10
CO-1704 Add labor chagnge job update

tsoliangwu0130:


src/main/groovy/edu/oregonstate/mist/personsapi/PersonsResource.groovy
```diff
@@ -377,8 +377,22 @@ class PersonsResource extends Resource {
...
+                // else {
+                //     logger.info("Updating $employmentType job")
+                //     dbFunctionOutput = bannerPersonsWriteDAO.updateJob(osuID, job)
+                //         .getString(BannerPersonsWriteDAO.outParameter)
+                // }
+            } else {
+                switch (employmentType) {
                 case studentEmploymentType:
+                    // TODO: This must be removed once the updateJob SP is completed.
                     if (update) {
```
Ruefa: I see the comment saying this will be removed but I don't think its reachable now anyways. If `update` is true then we won't be in this `else`

src/main/groovy/edu/oregonstate/mist/personsapi/PersonsResource.groovy
```diff
@@ -390,6 +404,7 @@ class PersonsResource extends Resource {
                     }
                     break
                 case graduateEmploymentType:
+                    // TODO: This must be removed once the updateJob SP is completed.
                     if (update) {
```
Ruefa: same as above

src/main/groovy/edu/oregonstate/mist/personsapi/PersonsResource.groovy
tsoliangwu0130: Ah, good catch, will fix it.

PR #85, opened 2020-02-17
CO-1713 openapi redesign

Ruefa:


openapi.yaml
```diff
@@ -17,1237 +21,222 @@ produces:
...
-            example: '97331'
-          countyCode:
+            enum: [FN, N, R, S, C]
+            description: |
+              Person's citizen code
+              - 'FN' - Foreign National
+              - 'N' - Non Resident Alien
+              - 'R' - Resident Alien
+              - 'S' - Substantial Presence Alien
+              - 'C' - Citizen
```
tsoliangwu0130: I checked the [openapi file](https://github.com/osu-mist/students-api/blob/master/openapi.yaml#L88-L142) of Students API and looks like we are using this format:

```yaml
- `code` - description
```

Can you change the codes here and follow the same style? Also, can you add a section in our style guild for this as well?

openapi.yaml
```diff
@@ -17,1237 +21,222 @@ produces:
...
             type: string
-            example: '41003'
-          nationCode:
+            description: |
+              Person's sex code
+              - 'M' - Male
+              - 'F' - Female
+              - 'N' - Non-specified
+          confidential:
             type: string
```
tsoliangwu0130: Should this be boolean? Also, I guess I'll rename it to maybe something like `isConfidential` or `confidentialInd`.

openapi.yaml
```diff
@@ -17,1237 +21,222 @@ produces:
...
-      data:
-        type: array
-        items:
-          $ref: '#/definitions/AddressResourceObject'
-  AddressResultObject:
+            description: If true, the person has opted in to being required to be listed as confidential
+          currentStudent:
+            type: boolean
+            description: True if person is a current student
+          currentEmployee:
```
tsoliangwu0130: Same naming issue with these two.

openapi.yaml
```diff
@@ -17,1237 +21,222 @@ produces:
...
+            type: boolean
+            description: True if person is a current employee
+          employeeStatus:
+            type: boolean
+            enum: ['A', 'B', 'L', 'F', 'P', 'T']
```
tsoliangwu0130: No need to quote for these items.

openapi.yaml
```diff
@@ -17,1237 +21,222 @@ produces:
...
-            type: string
-            description: 9 digits SSN
-            example: '123456789'
-          type:
-            $ref: '#/definitions/SSNType'
-          links:
-            $ref: '#/definitions/SelfLink'
-          attributes:
+            - firstName
+            - birthDate
```
tsoliangwu0130: I think `lastName` should be required as well.

openapi.yaml
```diff
@@ -17,1237 +21,222 @@ produces:
...
-          lastModified:
-            type: string
-            description: The date of the most recent change made to this record.
-            example: '2015-10-19'
+              firstName: *firstName
+              lastName: *lastName
+              middleName: *middleName
+              birthDay: *birthDay
+              citizen: *citizen
+              sex: *sex
```
tsoliangwu0130: Instead of referring each filed here, what do you think if you define a shared object called `commonPersonObject` or something else?

openapi.yaml
```diff
@@ -17,1237 +21,222 @@ produces:
...
-  AddressResultObject:
+            description: If true, the person has opted in to being required to be listed as confidential
+          currentStudent:
+            type: boolean
+            description: True if person is a current student
+          currentEmployee:
+            type: boolean
+            description: True if person is a current employee
+          employeeStatus:
+            type: boolean
```
tsoliangwu0130: Should be string.

openapi.yaml
```diff
@@ -17,1237 +21,222 @@ produces:
...
-          addressLine4:
+            description: ONID username
+          email:
             type: string
-          houseNumber:
+            description: Primary email address
+          birthDate: &birthDay
             type: string
-          city:
+            description: Birthday of person
```
tsoliangwu0130: Can you also add the format in description and an example? https://github.com/osu-mist/students-api/blob/9ab6cf63b26afc7802354ea4ef92ae56a12c1220/openapi.yaml#L735-L738

openapi.yaml
```diff
@@ -17,1237 +21,222 @@ produces:
...
-        example: meal-plans
+        $ref: '#/definitions/PersonType'
       links:
-          $ref: '#/definitions/SelfLink'
+        $ref: '#/definitions/SelfLink'
       attributes:
+        type: object
         properties:
-          mealPlan:
+          firstName: &firstName
```
tsoliangwu0130: Can you add examples to these fields if possible?

openapi.yaml
Ruefa: Yeah I thought about that but I really like having all the attributes defined in one place instead of spreading it out over multiple objects. Might be good to bring up in our next meeting.

tsoliangwu0130: Hmm, not sure did I misunderstand, but I mean you can still define them in one place and use `allOf` to merge when needed: https://github.com/osu-mist/express-api-skeleton/blob/86a6439fac923b34fbd936796fd3d161ba49eadc/openapi.yaml#L209-L211. So it will be something like:

```yaml
Resource:
    - $ref: '#/definitions/CommonPart'
    - properties:
        extraField1:
          type: string
        extraField2:
          type: string
```

Ruefa: In my code all of the person fields are defined under the `PersonResource` while in your example some are defined in `CommonPart` with some extra fields under `Resource`.  I just find it easier when I have to edit these fields in the future if they are all found in one place.

openapi.yaml
```diff
@@ -143,374 +82,417 @@ paths:
...
-                description: image/jpeg
-        404:
-          description: Person not found
-          schema:
-            $ref: '#/definitions/Error'
-  /persons/{osuID}/meal-plans:
+            Content-type:
+              type: string
+              description: image/jpeg
+        '404':
```
tsoliangwu0130: Also need a 400 and a 500.

openapi.yaml
```diff
@@ -143,374 +82,417 @@ paths:
...
     get:
       description: Get meal plans for a person
       tags:
         - meal-plans
       operationId: getMealPlans
       parameters:
-        - $ref: '#/parameters/osuID'
-        - $ref: '#/parameters/authorization'
+        - $ref: '#/parameters/osuId'
+        - name: mealPlan
```
tsoliangwu0130: Probably rename it to `mealPlanName` will be more clear.

openapi.yaml
```diff
@@ -143,374 +82,417 @@ paths:
...
+          in: query
+          type: string
+          required: false
+          description: Filter results by meal plan name.
```
tsoliangwu0130: I think it can be helpful if we can provide a paramter to filter meal plans by balances. What do you think?

openapi.yaml
```diff
@@ -143,374 +82,417 @@ paths:
...
-          description: Person or meal plan not found
-          schema:
-            $ref: '#/definitions/Error'
-  /persons/{osuID}/addresses:
+            $ref: '#/definitions/MealPlanResult'
+        '404':
+          $ref: '#/responses/404'
+        '500':
+          $ref: '#/responses/500'
+  /persons/{osuId}/address:
```
tsoliangwu0130: Should it be `addresses`?

openapi.yaml
```diff
@@ -143,374 +82,417 @@ paths:
...
           description: Successful response
           schema:
-            $ref: '#/definitions/MealPlansResultObject'
-        404:
-          description: Person not found
-          schema:
-            $ref: '#/definitions/Error'
-  /persons/{osuID}/meal-plans/{mealPlanID}:
+            $ref: '#/definitions/MealPlanSetResult'
+        '404':
```
tsoliangwu0130: We need a 400 for this method I think.

openapi.yaml
```diff
@@ -143,374 +82,417 @@ paths:
...
+          in: query
+          type: string
+          required: false
+          description: Filter results by state code.
+        - name: nation
+          in: query
+          type: string
+          required: false
+          description: Filter results by nation.
+        - name: nationCode
```
tsoliangwu0130: Can you add enum list to those filters by codes?

openapi.yaml
```diff
@@ -143,374 +82,417 @@ paths:
...
+        '200':
           description: Successful response
           schema:
-            $ref: '#/definitions/AddressesResultObject'
-        404:
-          description: Person not found
-          schema:
-            $ref: '#/definitions/Error'
+            $ref: '#/definitions/AddressSetResult'
+        '404':
```
tsoliangwu0130: We need 400 for this method.

openapi.yaml
```diff
@@ -143,374 +82,417 @@ paths:
...
+          $ref: '#/responses/404'
+        '500':
+          $ref: '#/responses/500'
     post:
       description: Create address for a person
       tags:
         - addresses
       operationId: postAddresses
```
tsoliangwu0130: For consistency, it should be `postAddress` since I think we can only create one address at a time.

openapi.yaml
```diff
@@ -143,374 +82,417 @@ paths:
...
           required: true
           schema:
-            $ref: '#/definitions/NewAddressResultObject'
-        - $ref: '#/parameters/osuID'
-        - $ref: '#/parameters/authorization'
+            $ref: '#/definitions/AddressPostBody'
+        - $ref: '#/parameters/osuId'
       responses:
-        200:
+        '200':
```
tsoliangwu0130: Not sure, but should we use 202 for the successful response here?

openapi.yaml
```diff
@@ -143,374 +82,417 @@ paths:
...
-            $ref: '#/definitions/ErrorArray'
-        404:
-          description: Person not found
-          schema:
-            $ref: '#/definitions/Error'
-  /persons/{osuID}/phones:
+            $ref: '#/definitions/AddressResult'
+        '400':
+          $ref: '#/responses/400'
+        '404':
```
tsoliangwu0130: A 409 may occur here as well.

openapi.yaml
```diff
@@ -143,374 +82,417 @@ paths:
...
       operationId: getPhones
       parameters:
+        - $ref: '#/parameters/osuId'
         - $ref: '#/parameters/addressType'
         - $ref: '#/parameters/phoneType'
-        - $ref: '#/parameters/osuID'
-        - $ref: '#/parameters/authorization'
+        - name: primary
+          in: query
+          type: string
```
tsoliangwu0130: Should this be a boolean?

openapi.yaml
```diff
@@ -143,374 +82,417 @@ paths:
...
-        400:
-          description: Bad request
-          schema:
-            $ref: '#/definitions/ErrorArray'
-        404:
-          description: Person not found
-          schema:
-            $ref: '#/definitions/Error'
+            $ref: '#/definitions/PhoneSetResult'
+        '404':
```
tsoliangwu0130: We need a 400 here as well.

openapi.yaml
```diff
@@ -143,374 +82,417 @@ paths:
...
+          $ref: '#/responses/404'
+        '500':
+          $ref: '#/responses/500'
     post:
-      description: Create phone numbers for a person
+      description: Create phone number for a person
       tags:
         - phones
       operationId: createPhones
```
tsoliangwu0130: Should be `postPhone`.

openapi.yaml
```diff
@@ -143,374 +82,417 @@ paths:
...
+            $ref: '#/definitions/PhonePostBody'
+        - $ref: '#/parameters/osuId'
+      responses:
+        '200':
+          description: Successful response
+          schema:
+            $ref: '#/definitions/PhoneResult'
+        '400':
+          $ref: '#/responses/400'
+        '404':
```
tsoliangwu0130: 409 here.

openapi.yaml
```diff
@@ -143,374 +82,417 @@ paths:
...
       responses:
-        200:
+        '200':
           description: Successful response
           schema:
-            $ref: '#/definitions/PhoneResultObject'
-        400:
-          description: Bad request
+            $ref: '#/definitions/EmailSetResult'
+        '404':
```
tsoliangwu0130: 400 may occur as well.

openapi.yaml
```diff
@@ -143,374 +82,417 @@ paths:
...
+          $ref: '#/responses/404'
+        '500':
+          $ref: '#/responses/500'
+    post:
```
tsoliangwu0130: Let's remove this method first. Since this method will insert new data (not supported in v1) into the database, we need to get approval from Joe or Kent before implement it.

openapi.yaml
```diff
@@ -143,374 +82,417 @@ paths:
...
 parameters:
-  osuID:
-    name: osuID
+  osuId:
+    name: osuId
     in: path
     required: true
-    description: OSU ID of person
+    description: OSU ID of a person
     type: string
```
tsoliangwu0130: Can you add pattern here? https://github.com/osu-mist/students-api/blob/master/openapi.yaml#L366

openapi.yaml
```diff
@@ -143,374 +82,417 @@ paths:
...
+    description: ID of meal plan
+    required: true
+    type: string
   employmentType:
     name: employmentType
     in: query
     description: The type of employment that the job record should be processed as.
     required: true
     type: string
     enum: [student, graduate]
```
tsoliangwu0130: This should go to job request body.

openapi.yaml
```diff
@@ -525,343 +507,331 @@ parameters:
...
+    schema:
+      $ref: '#/definitions/ErrorResult'
+securityDefinitions:
+  OAuth2:
+    type: oauth2
+    flow: application
+    tokenUrl: https://api.oregonstate.edu/oauth2/token
+    scopes:
+      full: Full access to the API
 definitions:
```
tsoliangwu0130: I think it'll be better to pause my review here for a while, otherwise my comments will be too many.

openapi.yaml
```diff
@@ -17,121 +21,56 @@ produces:
...
       tags:
         - persons
-      operationId: postPerson
+      operationId: getPersons
       parameters:
-        - name: personBody
-          in: body
-          description: Result object for a person
+        - in: body
+          name: body
```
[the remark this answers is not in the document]
Ruefa: I would prefer `personBody`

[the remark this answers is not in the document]
tsoliangwu0130: Keep in mind sometimes PATCH body and POST body can be different.

openapi.yaml
```diff
@@ -143,374 +82,417 @@ paths:
...
-        - name: jobBody
+        - name: body
           in: body
-          description: Result object for a job
           required: true
           schema:
-            $ref: '#/definitions/NewJobResultObject'
+            $ref: '#/definitions/JobPostBody'
+        - $ref: '#/parameters/osuId'
         - $ref: '#/parameters/employmentType'
```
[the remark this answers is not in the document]
Ruefa: Agreed

openapi.yaml
```diff
@@ -17,121 +21,56 @@ produces:
...
+            $ref: '#/definitions/PersonResult'
+        '404':
+          $ref: '#/responses/404'
+        '500':
+          $ref: '#/responses/500'
+  /persons/{osuId}/jobs:
     get:
       description: Get jobs associated with a person
       tags:
         - jobs
```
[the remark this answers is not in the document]
Ruefa: What is the value in doing this? It seems redundant to me since all these endpoints are already under the Students API.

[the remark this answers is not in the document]
tsoliangwu0130: You're probably right, most subresources endpoints are under the same resource (and it mostly are the API name), however, there are still some exceptions like locations api: https://github.com/osu-mist/locations-frontend-api/blob/develop/swagger.yaml
/services and /locations are two different base endpoints for this API. Just not sure what should we put in tags. Maybe we need to come up with a standard for all APIs.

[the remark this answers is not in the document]
Ruefa: I think in the event that you have multiple top level resources there is value to adding a tag for the resource. In the case where there is only one top resource it just adds bloat.

openapi.yaml
Ruefa: What can cause a 400 here?

tsoliangwu0130: If `width` is out of bound or not an integer.

openapi.yaml
Ruefa: That would work well with the `operation` filter parameter we can use now. I will also add a filter for `lastUsedLocation`.

openapi.yaml
Ruefa: Do we want to put all ~240 nation codes on there?

tsoliangwu0130: Shoot, hmm, I guess I personally will lean toward adding it to the API spec for the clearness. But I agree it's a giant list, so let's discuss this with other in the team meet and see how they feel.

openapi.yaml
Ruefa: It is 200 in v1. I'm not sure how to check what it should be.

tsoliangwu0130: All other POST and PATCH/PUT for this API are using 202 accept for successful responses because we need to call some SPs to process the data and it may take times until the data is really updated/inserted, I feel it'll be more consistent to use the same status code here as well. However, since this endpoint is actually calling an SP that won't need extra time to process, in other words, if we didn't get a thrown error, the new data is actually inserted. That's why 200 also sounds fair to me.

Ruefa: I agree I'll keep it 200

openapi.yaml
```diff
@@ -525,343 +605,340 @@ parameters:
...
+        example: 1
       salaryGroupCode:
         type: string
+        description: Salary group code for the job.
+        example: 12345A
       strsAssignmentCode:
         type: string
-      supervisorOsuID:
+        description: STRS assignment code for the job.
+        example: 12
```
tsoliangwu0130: If the value can be parsed as a number, it should be quoted: https://github.com/osu-mist/style-guides/blob/master/OpenAPI.md#quoting-conventions

openapi.yaml
```diff
@@ -525,343 +605,340 @@ parameters:
...
+      supervisorOsuId:
```
tsoliangwu0130: Can you also add a pattern for this?

openapi.yaml
```diff
@@ -525,343 +605,340 @@ parameters:
...
+        description: Organization code where the time sheet is sent for the job.
+        example: 123456
       timesheetOrganizationDesc:
         type: string
-        description: Organization title where time sheet is sent for this job.
+        description: Organization title where timesheet is sent for the job.
       timesheetOrganizationPredCode:
         type: string
         description: The next higher organizational code of the current timesheet organization.
+        example: 123456
```
tsoliangwu0130: This should be quoted as well.

openapi.yaml
```diff
@@ -525,343 +605,340 @@ parameters:
...
+        example: C12345
       supervisorSuffix:
         type: string
-        description: Suffix of the supervisor
+        description: Suffix of the supervisor.
       timesheetOrganizationCode:
         type: string
-        description: Organization code where time sheet is sent for this job.
+        description: Organization code where the time sheet is sent for the job.
+        example: 123456
```
tsoliangwu0130: This should be quoted as well.

openapi.yaml
```diff
@@ -525,343 +605,340 @@ parameters:
...
         type: string
         description: The next higher organizational code of the current timesheet organization.
+        example: 123456
       timesheetOrganizationPredDesc:
         type: string
         description: The next higher organizational title of the current timesheet organization.
       homeOrganizationCode:
         type: string
         description: The home organization code of the employee.
+        example: 123456
```
tsoliangwu0130: This should be quoted as well.

openapi.yaml
```diff
@@ -525,343 +605,340 @@ parameters:
...
       homeOrganizationDesc:
         type: string
         description: The home organization title of the employee.
       homeOrganizationPredCode:
         type: string
         description: The next higher organizational code of the current home organization.
+        example: 123456
```
tsoliangwu0130: This should be quoted as well.

openapi.yaml
```diff
@@ -872,59 +949,56 @@ definitions:
...
         description: Indicates the employee has consented to having their employment information shared.
-      salaryInformationRelease:
+      salaryInformationReleaseInd:
         type: boolean
-        description: Indicates the employee has consented to having their salary information shared.
+        description: Indicates the employee accepted or denied to have their salary information shared.
       salaryInformationReleaseDate:
         type: string
         format: date
         description: The date that the employee accepted or denied to have their salary information shared.
```
tsoliangwu0130: Can you add an example here? It's mainly for helping improve the UI of developer portal when we import this doc.

openapi.yaml
```diff
@@ -937,7 +1011,7 @@ definitions:
               type: string
               format: date
               description: Date this labor distribution becomes effective.
-              example: '2018-12-31'
+              example: 1970-1-1
```
tsoliangwu0130: Can you add pattern to this as well?

openapi.yaml
```diff
@@ -958,296 +1032,476 @@ definitions:
...
+      balance:
+        type: number
+        format: float
+        description: Remaining balance on meal plan in USD.
+        example: 3.50
+      lastUsedDate:
+        type: string
+        format: date-time
+        description: The last time the meal plan was used.
+        example: '2017-07-14T16:45:15Z'
```
tsoliangwu0130: I guess I will rename this to `lastUsedDateTime`. Also, do you think it will be handy if we have a filter for this? Something like `filter[lastUsedDate][before]=2017-07-14T16:45:15Z`? (I'm not sure for the operation though)

openapi.yaml
```diff
@@ -958,296 +1032,476 @@ definitions:
...
+      lastUsedLocation:
+        type: string
+        description: Description of the place the meal plan was last used.
+        example: 'Java II'
```
tsoliangwu0130: No need to quote.

openapi.yaml
```diff
@@ -164,353 +102,495 @@ paths:
...
-        - $ref: '#/parameters/authorization'
+        - $ref: '#/parameters/osuId'
+        - name: filter[mealPlanName]
+          in: query
+          type: string
+          required: false
+          description: Filter results by meal plan name.
+        - name: filter[balance]
+          in: query
+          type: string
```
tsoliangwu0130: I believe this is a number, and we will need operations for this parameter: https://github.com/osu-mist/style-guides/blob/master/OpenAPI.md#filter-query-parameters

openapi.yaml
```diff
@@ -958,296 +1032,476 @@ definitions:
...
+      addressType:
+        type: string
+        description: Code for the type of address record.
+        example: CM
+      addressLine1:
+        type: string
+        example: '123 Main Street'
+      addressLine2:
+        type: string
+        example: 'Apartment 4'
```
tsoliangwu0130: No need to quote.

openapi.yaml
```diff
@@ -958,296 +1032,476 @@ definitions:
...
+      - addressType
+      - city
+    properties:
+      addressType:
+        type: string
+        description: Code for the type of address record.
+        example: CM
+      addressLine1:
+        type: string
+        example: '123 Main Street'
```
tsoliangwu0130: No need to quote.

openapi.yaml
```diff
@@ -958,296 +1032,476 @@ definitions:
...
+        description: Human readable county.
+        example: Benton
+      nation:
+        type: string
+        description: Human readable nation/country name.
+        example: Albania
+      lastModified:
+        type: string
+        format: date
+        description: The date this address record was last modified.
```
tsoliangwu0130: Can you give an example and or format for this?

openapi.yaml
```diff
@@ -958,296 +1032,476 @@ definitions:
...
+      houseNumber:
+        type: string
+        example: '620'
+      city:
+        type: string
+        example: Corvallis
+      stateCode:
+        type: string
+        description: State code/abbreviation. This could also represent a province.
+        example: OR
```
tsoliangwu0130: I belive you have a enum list for this already.

openapi.yaml
```diff
@@ -958,296 +1032,476 @@ definitions:
...
+  AddressPostAttributes:
+    type: object
+    required:
+      - addressType
+      - city
+    properties:
+      addressType:
+        type: string
+        description: Code for the type of address record.
+        example: CM
```
tsoliangwu0130: I believe you have an enum list for this already.

openapi.yaml
```diff
@@ -958,296 +1032,476 @@ definitions:
...
+      lastUsedLocation:
+        type: string
+        description: Description of the place the meal plan was last used.
+        example: 'Java II'
+  MealPlanResource:
+    type: object
+    properties:
+      id: 
+        type: string
+        description: Meal plan ID
```
tsoliangwu0130: Can you add an example?

openapi.yaml
```diff
@@ -958,296 +1032,476 @@ definitions:
...
+        format: date
+        description: Date of last activity for this email
+        example: '1970-1-1'
+  EmailResource:
+    type: object
     properties:
       id:
         type: string
-        description: Globally unique ID for the phones.
+        description: Unique ID for an email address
```
tsoliangwu0130: What should it be? Can you give an example?

openapi.yaml
```diff
@@ -958,296 +1032,476 @@ definitions:
...
+        description: Describes the phone type enum.
+        example: Current
+      addressTypeDescription:
+        type: string
+        description: Describes the address type enum.
+        example: Current Mailing
+      lastModified:
+        type: string
+        description: The date of the most recent change made to this record.
+        example: '2015-10-19'
```
tsoliangwu0130: Can you add format for this?

openapi.yaml
```diff
@@ -958,296 +1032,476 @@ definitions:
...
+      emailAddress:
+        type: string
+        description: Email address for a person
+        example: john.smith@example.com
+      comment:
+        type: string
+        description: Comment associated with this email
+      preferredInd:
+        type: boolean
+        description: Indicator for a preferred email
```
tsoliangwu0130: I guess it's worthy to have a query parameter for this.

openapi.yaml
```diff
@@ -958,296 +1032,476 @@ definitions:
...
+          Type of email
+          - `STU` - Non-directory Student
+          - `ONID` - Student
+          - `VEND` - Vendor
+          - `VWEB` - Vendor
+          - `EMPL` - Employee
+          - `VPAY` - Vendor Pay
+          - `EMER` - Emergency Contact Persons Email for Student
+      emailAddress:
+        type: string
```
tsoliangwu0130: You can use `email` as format.

openapi.yaml
```diff
@@ -958,296 +1032,476 @@ definitions:
...
       type:
-        $ref: '#/definitions/PhoneType'
+        $ref: '#/definitions/EmailType'
+      links:
+        $ref: '#/definitions/SelfLink'
       attributes:
+        allOf:
+          - $ref: '#/definitions/EmailPostAttributes'
+          - $ref: '#/definitions/EmailAttributes'
+  EmailResult:
```
tsoliangwu0130: Are you using it anywhere?

openapi.yaml
```diff
@@ -958,296 +1032,476 @@ definitions:
...
+  EmailSetResult:
+    type: object
+    properties:
+      links:
+        $ref: '#/definitions/SelfLink'
+      data: 
+        type: array
+        items:
+          $ref: '#/definitions/EmailResource'
+  EmailPostBody:
```
tsoliangwu0130: Are you using it anywhere?

openapi.yaml
```diff
@@ -525,343 +605,340 @@ parameters:
...
       supervisorEmail:
         type: string
         description: The primary email of the supervisor for the job.
+        example: john.smith@example.com
       supervisorPositionNumber:
         type: string
-        description: Position number of the supervisor.
+        description: Position number for the supervisor.
+        example: C12345
       supervisorSuffix:
```
tsoliangwu0130: What do you think if we move these supervisor fileds under a supdervisor object? Will that be more clear?

openapi.yaml
```diff
@@ -525,343 +605,340 @@ parameters:
...
         description: First name of the supervisor for the job.
+        example: John
+      supervisorLastName:
+        type: string
+        description: Last name of the supervisor for the job.
+        example: Smith
       supervisorEmail:
         type: string
         description: The primary email of the supervisor for the job.
+        example: john.smith@example.com
```
tsoliangwu0130: You can use `email` for format here.

openapi.yaml
```diff
@@ -525,343 +605,340 @@ parameters:
...
+          - `S` - Substantial Presence Alien
+          - `C` - Citizen
+      sex:
+        type: string
+        description: |
+          Person's sex code
+          - `M` - Male
+          - `F` - Female
+          - `N` - Non-specified
+  PersonAttributes:
```
tsoliangwu0130: What do you think those fields I left in the comment of JIRA ticket?

* ethnicity
* marital status
* deceased status
* veteran status
* driver license
* etc.

openapi.yaml
Ruefa: I like that idea

openapi.yaml
```diff
@@ -164,353 +102,495 @@ paths:
...
+            - AP
+            - AE
+            - AA
+            - FC
+            - FR
+            - NU
+            - BY
+            - ENG
+          description: Filter results by state code.
+        - name: filter[nation]
```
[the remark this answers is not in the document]
Ruefa: From a users perspective they could have different uses and thus not be the same thing. If we return only one of the fields our users may have to convert them depending on their use cases.

openapi.yaml
```diff
@@ -164,353 +102,495 @@ paths:
...
+  mealPlanId:
+    name: mealPlanId
+    in: path
+    description: ID of meal plan
+    required: true
+    type: string
   employmentType:
-    name: employmentType
+    name: filter[employmentType]
     in: query
```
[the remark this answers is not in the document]
Ruefa: It was not but should be included on `GET /person/{osuId}/jobs`. I changed the description and set required to false to reflect its new use.

openapi.yaml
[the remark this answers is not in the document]
tsoliangwu0130: @Ruefa I agreed we should return both codes and descriptions, but I was not sure should we have two query parameters with the similar purposes. I think most users only use code search for most of our other APIs, that's why I thought we can reconsider the use case of name search parameters.

openapi.yaml
```diff
@@ -164,353 +102,495 @@ paths:
...
+    get:
+      description: Get emails for a person
+      tags:
+        - emails
+      operationId: getEmails
+      parameters:
+        - $ref: '#/parameters/osuId'
+        - name: filter[emailType]
+          in: query
+          type: string
```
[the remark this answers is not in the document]
Ruefa: good idea

openapi.yaml
```diff
@@ -525,343 +605,340 @@ parameters:
...
+        example: Adam
+      displayLastName:
+        type: string
+        description: Preferred last name
+        example: Smith
+      onid:
+        type: string
+        description: ONID username
+        example: smithj
+      email:
```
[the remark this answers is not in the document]
Ruefa: I think email is so commonly used that it should stay as a person attribute. The email endpoint can be for supplemental data.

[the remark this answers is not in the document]
tsoliangwu0130: But in this case which type of email should be returned here?

openapi.yaml
```diff
@@ -164,353 +102,495 @@ paths:
...
+  /persons/{osuId}/phones:
     get:
       description: Get phone numbers for a person
       tags:
         - phones
       operationId: getPhones
       parameters:
+        - $ref: '#/parameters/osuId'
         - $ref: '#/parameters/addressType'
         - $ref: '#/parameters/phoneType'
```
ianbrown9475: What's the difference between phoneType and addressType here? Do we need both?

openapi.yaml
```diff
@@ -164,353 +102,495 @@ paths:
...
+            - EMER
+          description: |
+            Type of email
+            - `STU` - Non-directory Student
+            - `ONID` - Student
+            - `VEND` - Vendor
+            - `VWEB` - Vendor
+            - `EMPL` - Employee
+            - `VPAY` - Vendor Pay
+            - `EMER` - Emergency Contact Persons Email for Student
```
ianbrown9475: Would we also want a filter for preferred email(s)?

openapi.yaml
tsoliangwu0130: @ianbrown9475 Yeah we need both of them. There are some overlaps between them, but there are still some are not. It's a very common use case that client needs to search by phoneType or addressType, so it'll be helpful if we have both of them implemented.

openapi.yaml
```diff
@@ -164,353 +102,495 @@ paths:
...
-                type: string
-                description: image/jpeg
-        404:
-          description: Person not found
-          schema:
-            $ref: '#/definitions/Error'
-  /persons/{osuID}/meal-plans:
+            Content-type:
+              type: string
+              description: image/jpeg
```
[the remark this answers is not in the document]
Ruefa: Can you leave an example? I'm unsure what you mean here.

openapi.yaml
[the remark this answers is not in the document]
Ruefa: Looks like we cant rely on preferred emails so I removed the field

openapi.yaml
```diff
@@ -958,296 +1032,476 @@ definitions:
...
+      links:
+        $ref: '#/definitions/SelfLink'
       data:
-        $ref: '#/definitions/NewAddressResourceObject'
-  NewAddressResourceObject:
+        $ref: '#/definitions/SsnResource'
+  MealPlanAttributes:
+    type: object
     properties:
+      mealPlan:
```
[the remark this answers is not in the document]
tsoliangwu0130: I agree with @ianbrown9475, or `mealPlanDescription` as you did for `jobDescription`.

openapi.yaml
```diff
@@ -937,317 +1058,539 @@ definitions:
...
+        type: number
+        format: float
+        description: Remaining balance on meal plan in USD.
+        example: 3.50
+      lastUsedDateTime:
+        type: string
+        format: date-time
+        description: The last time the meal plan was used.
+        example: '2017-07-14T16:45:15Z'
+      lastUsedLocation:
```
tsoliangwu0130: well... I'm not sure what should be combined and what should not now lol.

```json
{
    "lastUsed": {
        "dateTime": "2017-07-14T16:45:15Z",
        "location": "Java II"
    }
}
```

openapi.yaml
```diff
@@ -164,704 +128,833 @@ paths:
...
+          format: float
+          required: false
+          description: Filter results based on the remaining balance.
+        - name: filter[lastUsedDateTime][gt]
+          in: query
+          type: string
+          format: date-time
+          required: false
+          description: Filter results based on the last date and time the meal-plan was used.
+        - name: filter[lastUsedDateTime][gte]
```
tsoliangwu0130: Not sure, but I think we can remove `gt` and only keep `gte` and then  add `lte` as operations for these parameters.

openapi.yaml
```diff
@@ -937,317 +1058,539 @@ definitions:
...
+    type: object
+    required:
+      - addressType
+      - addressLine1
+      - city
+  AddressPostAttributes:
+    type: object
+    additionalProperties: false
+    properties:
+      addressType:
```
tsoliangwu0130: This can be a code-description object as well.

openapi.yaml
```diff
@@ -937,317 +1058,539 @@ definitions:
...
-                pattern: '\d{9}'
-  SSNResultObject:
+        type: array
+        items:  
+          $ref: '#/definitions/AddressResource'
+  AddressResult:
+    type: object
     properties:
-      links:
+      links:
```
tsoliangwu0130: Trailing spaces here.

openapi.yaml
```diff
@@ -164,704 +128,833 @@ paths:
...
+        type: string
+        description: ONID username
+        example: smithj
+      isConfidential:
+        type: boolean
+        description: If true, the person has opted in to being required to be listed as confidential
+      isCurrentStudent:
+        type: boolean
+        description: True if person is a current student
+      isCurrentEmployee:
```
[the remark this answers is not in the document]
Ruefa: Agreed. `isCurrentStudent` does the same thing. Think we should return `studentStatus` there as well?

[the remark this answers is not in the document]
tsoliangwu0130: I think `isCurrentStudent` can only be either`Y` and `N` (or maybe `null`) so it will make sense to keep it as boolean. We probably just wanna rename it to `currentStudentInd` for the consistency.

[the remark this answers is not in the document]
Ruefa:
```
New Student
Inactive Student
Withdrawn Student
Continuing
Readmit/Old Returning
Summer Only Student
Continuing Higher Education
Auditor
Non-matriculated student
```
There are other options.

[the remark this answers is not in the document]
Ruefa: This is already returned in students so we will keep it as is.

openapi.yaml
Ruefa: Yeah we should be careful but I think it works here.

openapi.yaml
```diff
@@ -164,704 +128,833 @@ paths:
...
+      employmentType:
+        type: string
+        enum: *employmentTypeEnums
+        description: Type of employment.
       effectiveDate:
         type: string
         format: date
         description: The date this job assignment is effective.
-        example: '2018-12-31'
+        example: '1970-1-21'
```
tsoliangwu0130: Should this be YYYY-MM-DD format?

openapi.yaml
```diff
@@ -164,704 +128,833 @@ paths:
...
+              - `O` - Overload
+          description:
+            type: string
+            description: Description of contract code
+      campus:
+        type: object
+        properties:
+          code:
+            type: string
+            description: Id of campus where this job will be performed.
```
tsoliangwu0130: For general sentences, I think ID is the correct form. `Id` is only used for the variables naming style.

openapi.yaml
Ruefa: Yup. I made that mistake in a lot of places and fixed them all.

openapi.yaml
```diff
@@ -164,1090 +135,1589 @@ paths:
...
+  osuId:
+    name: osuId
     in: path
     required: true
-    description: OSU ID of person
+    description: OSU ID of a person
     type: string
-  jobID:
-    name: jobID
+    pattern: '^\d{9}$'
```
tsoliangwu0130: Since we have pattern for this path parameter, we should return 400 if input doesn't match the regex.

openapi.yaml
```diff
@@ -164,1090 +135,1589 @@ paths:
...
+            collectionFormat: csv
+          required: false
+          enum: &emailTypeEnums
+            - ONID
+            - STU
+            - VEND
+            - VWEB
+            - EMPL
+            - VPAY
+            - EMER
```
ianbrown9475: I think this enum should be under `items` instead of under the array parameter

openapi.yaml
ianbrown9475: Didn't we decide that for id path parameters, if it doesn't match the pattern, we should still respond with a 404?

openapi.yaml
```diff
@@ -164,1090 +135,1589 @@ paths:
...
+        description: Extension for the phone number. Can be returned in different formats.
+        example: '4443333'
+        pattern: '^[0-9]{1,7}$'
+      phoneExtension:
+        type: string
+        description: Extension for the phone number. Can be returned in different formats.
+        example: '20'
+        pattern: '^[0-9]{1,4}$'
+      primaryInd:
+        type: b
```
ianbrown9475: It looks like these email post attributes are only used as part of the GET /.../emails endpoint. If there won't be a POST endpoint, could you just move these into EmailAttributes?

openapi.yaml
tsoliangwu0130: @ianbrown9475 You're right, I was looking into the wrong place, feel free to ignore my comment here.

openapi.yaml
Ruefa: Good catch. There was a post for email but it got removed.

PR #87, opened 2020-03-06
CO-1739 convert openapi 2 to 3

Ruefa:


openapi.yaml
```diff
@@ -31,71 +25,69 @@ paths:
...
-          type: string
-          enum: &jobStatusEnums
-            - A
-            - B
-            - L
-            - F
-            - P
-            - T
           required: false
           description: |
```
tsoliangwu0130: Seems like you use both `>` and `|`. I guess we should use `>` for most of the cases. Can you check the code and make them consistent and also create a PR in style guide for this?

openapi.yaml
```diff
@@ -469,1250 +491,1393 @@ paths:
...
+        application/json:
+          schema:
+            $ref: '#/components/schemas/ErrorResult'
+    '404':
+      description: Resource not found
+      content:
+        application/json:
+          schema:
+            $ref: '#/components/schemas/ErrorResult'
+    '409':
```
tsoliangwu0130: Looks like @ianbrown9475 mentioned this in the last PR: https://github.com/osu-mist/persons-api/pull/85#discussion_r384617963

Thoughts?

openapi.yaml
```diff
@@ -469,1250 +491,1393 @@ paths:
...
+          type: string
+          description: First name of person
+          example: John
+        lastName:
+          type: string
+          description: Last name of person
+          example: Smith
+        middleName:
+          type: string
+          nullable: true
```
tsoliangwu0130: Nice. Do you think we should also explicitly mark those required fields be `nullable: false`?

openapi.yaml
```diff
@@ -469,1250 +491,1393 @@ paths:
...
+                Person's citizen code
+                - `FN` - Foreign National
+                - `N` - Non Resident Alien
+                - `R` - Resident Alien
+                - `S` - Substantial Presence Alien
+                - `C` - Citizen
+        sex:
+          type: string
+          nullable: true
+          enum:
```
tsoliangwu0130: Hmm, not sure, but should the value of `enum` always be moved to the next line?

openapi.yaml
```diff
@@ -469,1250 +491,1393 @@ paths:
...
+            [M, F, N]
+          description: |
+            Person's sex code
+            - `M` - Male
+            - `F` - Female
+            - `N` - Non-specified
+    PersonAttributes:
```
tsoliangwu0130: I don't think any of these attributes can be null beside `displayMiddleName` for a good person data. Since we have `nullable` in openapi 3, it's handy for us to check if the return data is expected or not.

openapi.yaml
```diff
@@ -469,1250 +491,1393 @@ paths:
...
+          nullable: true
+          description: The primary email of the supervisor for the job.
+          example: john.smith@example.com
+        positionNumber:
+          type: string
+          description: Position number for the supervisor.
+          example: C12345
+        suffix:
+          type: string
+          nullable: true
```
tsoliangwu0130: I don't think this field can be null.

openapi.yaml
```diff
@@ -469,1250 +491,1393 @@ paths:
...
+              - effectiveDate
+              - accountIndexCode
+              - distributionPercent
+    JobPostAttributes:
+      type: object
+      additionalProperties: false
+      properties:
+        suffix:
+          type: string
+          nullable: true
```
tsoliangwu0130: Same, suffix can't be null.

openapi.yaml
```diff
@@ -469,1250 +491,1393 @@ paths:
...
+              description: The reason the last change was made to the job record.
+        fullTimeEquivalency:
+          type: number
+          nullable: true
+          format: float
+          description: Full time equivalency where full time = 1.
+          example: 0.25
+        appointmentPercent:
+          type: number
+          nullable: true
```
tsoliangwu0130: This can't be null.

openapi.yaml
```diff
@@ -469,1250 +491,1393 @@ paths:
...
+        changeReason:
+          type: object
+          nullable: true
+          properties:
+            code:
+              type: string
+              description: The reason the last change was made to the job record.
+        fullTimeEquivalency:
+          type: number
+          nullable: true
```
tsoliangwu0130: This can't be null. Can you check all the rest attributes and make sure they can or can't be null?

```SQL
SELECT COUNT(*) FROM TABLE_NAME WHERE COLUMN_NAME IS NULL;
```

openapi.yaml
Ruefa: `>` strips linebreaks while `|` does not so they have different uses. I can still add this to the style guide.

openapi.yaml
Ruefa: I missed that. Its in now.

openapi.yaml
Ruefa: It throws errors if you try that

tsoliangwu0130: Really? This works for me:

```yaml
enum: &statusCodeEnums [A, B, L, F, P, T]
```

Ruefa: You're right. Looks like its just a bug with the syntax checker im using.

openapi.yaml
Ruefa: Not sure. Doesn't look like its worth the effort to me.

tsoliangwu0130: We can definitely have our integration test or validation logic only valid the data type only if `nullable` is set to `true`. I'm thinking about this from the UI wise. As the screenshot attached, the `nullable` attribute will only show if you specify it.

Screen Shot 2020-03-11 at 10 01 34 AM

There's a red star specified required fields so it should be fine for request body, however, it can be a little confusing in response schema.

Screen Shot 2020-03-11 at 10 06 32 AM

Ruefa: I added nullable to all fields

PR #89, opened 2020-03-20
CO-1762 implement get images

Ruefa:


src/api-routes/persons/{osuId}/images.js
```diff
@@ -0,0 +1,42 @@
+import sharp from 'sharp';
+
+import { errorHandler, errorBuilder } from 'errors/errors';
+import { getImageById } from 'db/oracledb/images-dao';
```
tsoliangwu0130: I think this one should be moved above.

openapi.yaml
```diff
@@ -588,6 +579,16 @@ components:
       required: true
       schema:
         type: string
+    imageWidth:
+      in: query
+      name: filter[width]
```
tsoliangwu0130: Hmm, not sure, but I think this parameter is not a "filter". I'm fine naming this for the conssitency, just curious your thoughts.

src/api-routes/persons/{osuId}/images.js
```diff
@@ -0,0 +1,42 @@
...
+    if (!image) {
+      return errorBuilder(res, 404, 'No image with the specified OSU ID was not found.');
+    }
+
+    res.contentType('image/jpeg');
+    let result;
+    const sharpImage = sharp(image);
+    if (width) {
+      result = await sharpImage.resize(width, null).toBuffer();
+    } else if (await sharpImage.metadata().width > maxWidth) {
```
tsoliangwu0130: Will we ever run into this? I think the express-openapi package has validated this parameter and will return a 400 if the input if over the maximum.

src/db/oracledb/images-dao.js
```diff
@@ -0,0 +1,28 @@
...
+ * @param {string} osuId OSU ID of person to select
+ * @returns {Promise<object>} Serialized person resource from person-serializer
+ */
+const getImageById = async (osuId) => {
+  const connection = await getConnection();
+  try {
+    const query = { osuId };
+    const { rows } = await connection.execute(contrib.getImageById(), query);
+
+    if (!rows || _.isEmpty(rows)) {
```
tsoliangwu0130: We should return a [default image](https://github.com/osu-mist/persons-api/blob/master/src/main/resources/defaultImage.jpg) if image not found (but person exists) instead of a 404 since the resource should be person.

src/api-routes/persons/{osuId}/images.js
Ruefa: I added this for the case where someone has an image that is by default greater than the max width. Not sure how much this matters but the size of the response starts to balloon when you get passed 2000 pixel width.

openapi.yaml
Ruefa: I checked out Json API specification and only filtering parameters get to be filter[] params. I set it back to just `width`.

src/api-routes/persons/{osuId}/images.js
```diff
@@ -15,14 +15,20 @@ const { maximum: maxWidth } = openapi.components.parameters.imageWidth.schema;
...
   try {
     const { osuId } = req.params;
-    const { width } = parseQuery(req.query);
+    const { width } = req.query;
     const image = await getImageById(osuId);
 
+    // return default image if no image is returned from data source
     if (!image) {
-      return errorBuilder(res, 404, 'No image with the specified OSU ID was not found.');
+      return fs.readFile('src/resources/defaultImage.jpg', (err, data) => {
```
tsoliangwu0130: Looks like `width` parameter doesn't work on default image. Can you take a look?

Ruefa: It doesn't work by design but I can change that.

PR #90, opened 2020-03-23
CO-1763 implement get addresses

Ruefa:


src/serializers/addresses-serializer.js
```diff
@@ -0,0 +1,75 @@
...
+      const splitKey = key.split('.');
+      if (splitKey.length > 1) {
+        if (!address[splitKey[0]]) {
+          address[splitKey[0]] = {};
+        }
+        address[splitKey[0]][splitKey[1]] = value;
+      }
+    });
+  });
+};
```
Ruefa: Handling sub-objects in the jobs endpoint took a while. This is one way it can be done programmatically if you name the fields in the contrib correctly. What are your thoughts? If you like it we can improve on it and add it to the skeleton.

src/api-routes/persons/{osuId}/addresses.js
```diff
@@ -0,0 +1,31 @@
...
+
+/**
+ * Get addresses by OSU ID
+ *
+ * @type {RequestHandler}
+ */
+const get = async (req, res) => {
+  try {
+    const { osuId } = req.params;
+    const { query } = req;
```
ianbrown9475: I think you could also do `const { query, params: { osuId } } = req;`

src/db/oracledb/addresses-dao.js
```diff
@@ -0,0 +1,38 @@
...
+    parsedQuery.stateCode = parsedQuery.stateCode ? parsedQuery.stateCode : null;
+    parsedQuery.nationCode = parsedQuery.nationCode ? parsedQuery.nationCode : null;
+
+    const { rows } = await connection.execute(contrib.getAdresses(), parsedQuery);
+
+    if (_.isEmpty(rows)) {
+      return undefined;
+    }
+
+    return serializeAddresses(rows, query, osuId);
```
ianbrown9475: Instead of having the DAO return the result of the serializer, I think it should return the raw data. Then, the path handler can pass the result to the serializer. This will help us avoid things like when we want to pass arguments to the serializer that the DAO doesn't need. It also modularizes things a little more and makes the testing a little simpler too.

src/db/oracledb/addresses-dao.js
```diff
@@ -0,0 +1,38 @@
...
+const getAddressesById = async (osuId, query) => {
+  const connection = await getConnection();
+  try {
+    const parsedQuery = parseQuery(query);
+    parsedQuery.osuId = osuId;
+    parsedQuery.addressType = parsedQuery.addressType ? parsedQuery.addressType : null;
+    parsedQuery.city = parsedQuery.city ? parsedQuery.city : null;
+    parsedQuery.county = parsedQuery.county ? parsedQuery.county : null;
+    parsedQuery.stateCode = parsedQuery.stateCode ? parsedQuery.stateCode : null;
+    parsedQuery.nationCode = parsedQuery.nationCode ? parsedQuery.nationCode : null;
```
ianbrown9475: This logic is repeated for all of these properties, so I think it can be refactored.

src/db/oracledb/addresses-dao.js
```diff
@@ -0,0 +1,38 @@
...
+import { contrib } from './contrib/contrib';
+
+/**
+ * Queries data source for raw address data and passes it to the serializer before returning
+ *
+ * @param {string} osuId OSU ID of person to select addresses from
+ * @param {object} query Query parameters passed in with request
+ * @returns {Promise<object>} Serialized address resource
+ */
+const getAddressesById = async (osuId, query) => {
```
ianbrown9475: Not sure about this, but would it be better to call this `getAddressesByOsuId` instead to be more specific?

src/serializers/addresses-serializer.js
```diff
@@ -0,0 +1,75 @@
...
+  _.forEach(rawAddresses, (address) => {
+    _.forEach(address, (value, key) => {
+      const splitKey = key.split('.');
+      if (splitKey.length > 1) {
+        if (!address[splitKey[0]]) {
+          address[splitKey[0]] = {};
+        }
+        address[splitKey[0]][splitKey[1]] = value;
+      }
+    });
```
ianbrown9475: I think in general, when we're taking an object and then modifying both keys and values, we should use `_.transform` or `_.reduce`. I'm not sure if the values are just being moved to different keys or if some of them are kept how they are. If they're all being moved, `_.mapKeys` might work better. Not sure about here though since depth is added.

Actually, if the purpose of this function is just to evaluate `prop1.prop2: val` as `{ prop1: { prop2: val } }`, you could look into other lodash methods like `_.set`. Many of them can evaluate nested paths like this.

src/serializers/addresses-serializer.js
ianbrown9475: Agreed, I think that format makes the most sense.

src/api-routes/persons/{osuId}/addresses.js
```diff
@@ -0,0 +1,31 @@
...
+    const result = await getAddressesById(osuId, query);
+    if (!result) {
```
tsoliangwu0130: Same thing as images, result not found can possible be "the person exist but he doesn't have address or the filtered address doesn't exist". In this case, you need to return a 200 empty array instead of a 404.

You can reproduce the issue by searching `/persons/:osuId/addresses?filter[city]=foo`

src/api-routes/persons/{osuId}/addresses.js
```diff
@@ -0,0 +1,29 @@
...
+/**
+ * Get addresses by OSU ID
+ *
+ * @type {RequestHandler}
+ */
+const get = async (req, res) => {
+  try {
+    const { query, params: { osuId } } = req;
+    const result = await getAddressesByOsuId(osuId, query);
+    const serializedAddresses = serializeAddresses(result, query, osuId);
```
tsoliangwu0130: We only wanna return an empty array if the person exists but no address records are found. If a person doesn't exist, we should return 404.

You can execute a query to do a simple check before running the real query like this: https://github.sig.oregonstate.edu/ecs-data/persons-api-contrib/blob/2cab8153eb93a18d35ecbd22f51c18668ae7d12b/AbstractPersonsDAO.groovy#L4-L9

Same thing for all other sub-resource endpoints.

src/utils/format-sub-objects.js
```diff
@@ -0,0 +1,42 @@
...
+ * { 'addressType.code': 'CM' } => { addressType: { code: 'CM' } }
+ *
+ * @param {object[]} rawData raw output from data source
+ */
+const formatSubObjects = (rawData) => {
+  _.forEach(rawData, (data) => {
+    _.forEach(data, (value, key) => {
+      const splitKey = key.split('.');
+      if (splitKey.length > 1) {
+        expandKey(splitKey, data, value);
```
tsoliangwu0130: Should you use `_.set()`?

PR #91, opened 2020-03-26
CO-1764 implement get phones

Ruefa:


src/api-routes/persons/{osuId}/phones.js
```diff
@@ -0,0 +1,36 @@
+import { personExists } from 'db/oracledb/persons-dao';
+import { serializePhones } from 'serializers/phones-serializer';
+import { errorHandler, errorBuilder } from 'errors/errors';
+import { getPhones } from 'db/oracledb/phones-dao';
```
tsoliangwu0130: Can you fix the order here?

src/api-routes/persons/{osuId}/addresses.js
```diff
@@ -12,11 +12,12 @@ const get = async (req, res) => {
   try {
     const { query, params: { osuId } } = req;
 
-    if (!await personExists(osuId)) {
+    const pidm = await personExists(osuId);
```
tsoliangwu0130: We probably shouldn't use `pidm` as variable name since it's also Ellucian defined column name. I'd recommend changing it to  something like `personId` or `internalId`.

src/db/oracledb/phones-dao.js
```diff
@@ -1,20 +1,24 @@
...
- * @param {string} osuId OSU ID of person to select
+ * @param {string} pidm PIDM of person to select
  * @param {object} query query parameters passed in with the request
  * @returns {Promise<object>} Raw phone data from data source
  */
-const getPhones = async (osuId, query) => {
+const getPhones = async (pidm, query) => {
   const connection = await getConnection();
   try {
     const parsedQuery = parseQuery(query);
```
tsoliangwu0130: `filter[primaryInd]=foo` is considered as a valid parameter which shouldn't be. We need to add validation logic to only allow boolean.

src/api-routes/persons/{osuId}/addresses.js
Ruefa: I'll make the change but not sure why it being Ellucian defined is a problem. Is it supposed to be private info or something?

tsoliangwu0130: Yeah, based on the contract, we shouldn't expose table name and column name outside. Just using pidm should probably be fine since it's not the full column name, but I guess naming it implicitly will be safer.

src/api-routes/persons/{osuId}/images.js
```diff
@@ -1,9 +1,10 @@
 import fs from 'fs';
 import sharp from 'sharp';
 
-import { getImageById } from 'db/oracledb/images-dao';
 import { errorHandler } from 'errors/errors';
+import { getImageById } from 'db/oracledb/images-dao';
 import { openapi } from 'utils/load-openapi';
+import { personExists } from 'db/oracledb/persons-dao';
```
tsoliangwu0130, line 7: Can you fix the order here?

PR #95, opened 2020-04-07
CO-1769 implement get emails

Ruefa:


src/serializers/emails-serializer.js
```diff
@@ -0,0 +1,66 @@
...
+
+/**
+ * Some fields need to be prepared for the serializer
+ *
+ * @param {object[]} rawEmails raw data from data source
+ */
+const prepareRawEmails = (rawEmails) => {
+  _.forEach(rawEmails, (email) => {
+    email.preferredInd = email.preferredInd === 'Y';
+  });
```
tsoliangwu0130, line 22: Seems like `lastActivityDate` is not using the format defined here: https://github.com/osu-mist/persons-api/blob/0792a3fc0212dd25f7c6f4be4df4198719d660f6/openapi.yaml#L1920

PR #97, opened 2020-04-15
CO-1776 implement get meal plans

Ruefa:


config/default-example.yaml
```diff
@@ -15,9 +15,15 @@ authentication:
...
   dataSources: ['oracledb']
   oracledb:
-    connectString: ${DB_URL}
-    user: ${DB_USER}
-    password: ${DB_PASSWD}
     poolMin: 4
     poolMax: 4
     poolIncrement: 0
+    oracleSources: ['bannerRead', 'odsRead']
+    bannerRead:
```
tsoliangwu0130, line 22: As we discussed, I think we can just use `banner` and `ods` since banner read and banner write will be the same instance in v2.

config/default-example.yaml
```diff
@@ -15,9 +15,15 @@ authentication:
...
+      connectString: ${DB_URL}
+      user: ${DB_USER}
+      password: ${DB_PASSWD}
+    odsRead:
+      connectString: ${DB_URL}
+      user: ${DB_USER}
+      password: ${DB_PASSWD}
```
tsoliangwu0130, line 29: These env vars are named exactly the same as `bannerRead` which will cause the issue. Can you name them differently? Take a look at our env repo as an example.

src/db/oracledb/connection.js
```diff
@@ -16,7 +16,7 @@ const threadPoolSize = dbConfig.poolMax + (dbConfig.poolMax / 5);
 process.env.UV_THREADPOOL_SIZE = threadPoolSize > 128 ? 128 : threadPoolSize;
 
 /** Connection pool */
```
tsoliangwu0130, line 18: Connection pools.

src/db/oracledb/connection.js
```diff
@@ -26,19 +26,24 @@ let pool;
...
 
 /**
  * Get a connection from a created pool. Creates pool if it hasn't been created yet.
  *
+ * @param {string} pool Name of the connection pool to use
  * @returns {Promise} Promise object represents a connection from created pool
  */
-const getConnection = async () => {
-  if (!pool) {
+const getConnection = async (pool = 'bannerRead') => {
```
tsoliangwu0130, line 42: Why do we need a default value for `pool`? I think explicitly assign value to this var is clearer plus we should throw errors if pool name is not recoginized.

src/db/oracledb/meal-plans-dao.js
```diff
@@ -0,0 +1,38 @@
...
+
+    // create query string
+    const getMealPlansQuery = contrib.getMealPlansByOsuId(parsedQuery);
+    // strip operators from parsedQuery so they are not passed into the query execution
+    _.forEach(parsedQuery, (value, param) => {
+      if (value.operator) {
+        parsedQuery[param] = value.value;
+      }
+    });
+    const { rows } = await connection.execute(
```
tsoliangwu0130, line 28: I think it's not long enough to break into three lines.

src/db/oracledb/connection.js
```diff
@@ -54,7 +56,7 @@ const getConnection = async (pool = 'bannerRead') => {
 const validateOracleDb = async () => {
   let connection;
   try {
-    connection = await getConnection();
```
tsoliangwu0130, line 52: I think we should also validate ODS.

src/tests/unit/connection.test.js
```diff
@@ -20,7 +20,7 @@ describe('Test oracledb connection module', () => {
   beforeEach(() => {
     configGetStub = sinon.stub(config, 'get')
       .withArgs('dataSources.oracledb')
-      .returns({});
+      .returns({ oracleSources: ['banner'], banner: {} });
```
tsoliangwu0130, line 23: I'd recommend using a fake instance name like `fakeDb` or something similar for testing. By doing this, it'll be guaranteed we stub all necessary functions while testing, instead of passing the tests by coincidence.

Ruefa: Nice catch that was a bug

src/db/oracledb/connection.js
```diff
@@ -54,18 +54,20 @@ const getConnection = async (pool) => {
...
-    if (connection) {
-      await connection.close();
+  await _.asyncEach(dbConfig.oracleSources, async (source) => {
+    let connection;
+    try {
+      connection = await getConnection(source);
+      await connection.execute('SELECT 1 FROM DUAL');
+    } catch (err) {
+      logger.error(err);
+      throw new Error('Unable to connect to Oracle database');
```
tsoliangwu0130, line 64: A small comment, since we'll validate more than one db for this API, can you log the db name if the error is thrown?

PR #98, opened 2020-04-22
CO-1779 implement get meal plans by ID

Ruefa:


src/api-routes/persons/{osuId}/meal-plans/{mealPlanId}.js
```diff
@@ -0,0 +1,31 @@
+import { errorHandler, errorBuilder } from 'errors/errors';
+import { getMealPlanByMealPlanId } from 'db/oracledb/meal-plans-dao';
+import { serializeMealPlan } from 'serializers/meal-plans-serializer';
+import { personExists } from 'db/oracledb/persons-dao';
```
tsoliangwu0130, line 4: Some orders need to be fixed.

src/db/oracledb/meal-plans-dao.js
```diff
@@ -32,4 +32,27 @@ const getMealPlansByOsuId = async (osuId, query) => {
...
+  const connection = await getConnection('ods');
+  try {
+    const query = { osuId, mealPlanId };
+    const { rows } = await connection.execute(contrib.getMealPlansByOsuId(query), query);
+
+    if (rows.length < 1) {
+      return undefined;
+    }
+
+    return rows[0];
```
tsoliangwu0130, line 52: Two ways to improved:

```js
return rows.length < 1 ? undefined : rows[0];
```

or just return `rows[0]` if `row.length >= 1`. Return undefined if not necessary.

```js
if (rows.length >= 1) {
  return rows[0];
}
```

I personally prefer the first one since the second one is too implicitly to me.

Ruefa: I agree the first one is better

PR #102, opened 2020-05-05
CO-1796 implement post phones

Ruefa:


src/api-routes/persons/{osuId}/phones.js
```diff
@@ -1,7 +1,7 @@
 import { errorHandler, errorBuilder } from 'errors/errors';
```
tsoliangwu0130, line 1: Import order here should be fixed.

src/db/oracledb/phones-dao.js
```diff
@@ -1,29 +1,122 @@
...
+
+    if (body.phoneType !== body.addressType) {
+      const phoneAddressType = await phoneHasSameAddressType(
+        connection,
+        internalId,
+        body.addressType,
+      );
+      if (phoneAddressType.primaryInd) {
+        return new Error(
+          'A primary phone record with the $addressType address code already exists',
```
tsoliangwu0130, line 90: I think this should be a template string?

Ruefa: lol copied from v1 and forgot to change it

PR #103, opened 2020-05-05
CO-1756 implement post person

Ruefa:


src/api-routes/persons.js
```diff
@@ -11,6 +11,10 @@ const post = async (req, res) => {
   try {
     const { body: { data: { attributes } } } = req;
 
+    if (new Date(attributes.birthDate) > new Date()) {
+      return errorBuilder(res, 400, ["birthDate can't be a future date"]);
```
tsoliangwu0130, line 15: Single quote.

src/api-routes/persons.js
```diff
@@ -11,6 +11,10 @@ const post = async (req, res) => {
   try {
     const { body: { data: { attributes } } } = req;
 
+    if (new Date(attributes.birthDate) > new Date()) {
```
tsoliangwu0130, line 14: Since we have `moment` installed in the skeleton, I personally recommend using it for all time/date related codes since it's more clear in my opinion: https://momentjs.com/docs/

E.g. You can use `moment#isAfter` in this case.

src/api-routes/persons.js
Ruefa: Using double quotes because of the single quote in the string

tsoliangwu0130: Oops, I missed it. Nevermind then.

PR #104, opened 2020-05-06
CO-1809 Add CLASSIFIED_IND

tsoliangwu0130:


src/main/groovy/edu/oregonstate/mist/personsapi/mapper/JobsMapper.groovy
```diff
@@ -9,6 +9,23 @@ import java.sql.SQLException
...
+        if (classificationCode.startsWith('C') || classificationCode.startsWith('G')) {
+            classification = 'Classified'
+        } else if (classificationCode.startsWith('T')) {
+            classification = 'Temporary'
+        } else if (classificationCode.startsWith('X')) {
+            classification = 'Student'
+        } else if (classificationCode.startsWith('U')) {
+            classification = 'Unclassified'
+        } else if (classificationCode == 'XX') {
+            classification = 'Unpaid Appt'
```
Ruefa, line 24: You can never get the `classificationCode == 'XX'` case because `classificationCode.startsWith('X')` will pass first.

tsoliangwu0130: Oops, nice catch, fixed.

PR #112, opened 2020-05-26
CO-1788 implement post jobs

Ruefa:


src/db/oracledb/jobs-dao.js
```diff
@@ -37,23 +120,316 @@ const getJobs = async (internalId, query) => {
...
+    contrib.validateChangeReasonCode(),
+    { changeReasonCode },
+  );
+  return rows[0].count > 0;
+};
+
+/**
+ * Determines which Epaf to execute based on fields in body
+ *
+ * @param {boolean} update True if updating record. False if creating record
```
tsoliangwu0130, line 382: Just my two cents, I think using an enum string as controller (`update` and `create`) like is clearer than using boolean here. We can also easily extend in the future if there's a thrid different case we need to handle.

src/db/oracledb/jobs-dao.js
```diff
@@ -37,23 +120,316 @@ const getJobs = async (internalId, query) => {
...
+  const connection = await getConnection('banner');
+  try {
+    let result;
+    const { studentEmployeeInd, changeReason: { code: changeReasonCode } } = body;
+    const employmentType = studentEmployeeInd ? 'student' : 'graduate';
 
-    if (rows.length > 1) {
-      throw new Error(`Multiple job records found for job ID ${jobId}`);
+    if (employmentType === 'student' && !/^C5[0-2]/.test(body.positionNumber)) {
+      return new Error('Position number for students must start with C50, C51, or C52');
```
tsoliangwu0130, line 396: Interesting, I don't think we have this in v1?

src/db/oracledb/jobs-dao.js
```diff
@@ -37,23 +120,316 @@ const getJobs = async (internalId, query) => {
...
 
-    return rows[0];
+      if (update) {
+        if (changeReasonCode === 'NONE') {
+          result = await updateLaborChangeJob(connection, osuId, body);
+        } else if (changeReasonCode === 'BREAP') {
+          if (employmentType === 'student') {
+            result = await studentUpdateJob(connection, osuId, body);
+          } else if (employmentType === 'graduate') {
+            result = await graduateUpdateJob(connection, osuId, body);
```
tsoliangwu0130, line 413: `studentUpdateJob` and `graduateUpdateJob` are almost the same besides the contrib they call right? We can probably combine them into one `updateJob` and pass the `employmentType` as one argument to determine which contrib should be used. Same thing for `createJob`.

src/db/oracledb/jobs-dao.js
```diff
@@ -37,23 +120,316 @@ const getJobs = async (internalId, query) => {
...
+        } else {
+          await updateJob(connection, osuId, body);
+        }
+      } else if (employmentType === 'student') {
+        result = await studentCreateJob(connection, osuId, body);
+      } else if (employmentType === 'graduate') {
+        result = await graduateCreateJob(connection, osuId, body);
+      }
+    }
+    // null === success
```
tsoliangwu0130, line 424: What's this?

openapi.yaml
```diff
@@ -1162,14 +1123,14 @@ components:
...
+        employeeGroup:
           type: object
           properties:
             code:
               type: string
-              nullable: false
-              description: Employer identification code.
-              example: ORST
+              nullable: true
+              enum: [NOSU]
```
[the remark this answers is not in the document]
Ruefa: Null is also accepted and more values may be added in the future. See Illya's comment https://xe.ucsadm.oregonstate.edu:9990/ords/osprod_actwon/actwon.act_servicerequest_main.view_ticket?pSR=59118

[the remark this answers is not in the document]
tsoliangwu0130: Ok, so is it only for `POST` or also for `PATCH`?
Right now since you joint `JobPostAttributes` and `JobAttributes` for `JobResource` in [openapi](https://github.com/osu-mist/persons-api/blob/f014c16b7da2118ceae8b47a0e945cbc6a2feaa0/openapi.yaml#L1476-L1477), it makes this field in GET jobs body also only allows `NOSU` and null, which I think is not correct.

src/db/oracledb/jobs-dao.js
```diff
@@ -37,23 +120,316 @@ const getJobs = async (internalId, query) => {
...
+ * @param {string} jobId ID of a job record
+ * @returns {object} Single raw job record
+ */
 const getJobByJobId = async (internalId, jobId) => {
   const connection = await getConnection('banner');
   try {
-    const [positionNumber, suffix] = jobId.split('-');
-    const binds = { internalId, positionNumber, suffix };
-    const { rows } = await connection.execute(contrib.getJobs(binds), binds);
+    return await getJobByJobIdWithConnection(connection, internalId, jobId);
```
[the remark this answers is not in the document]
Ruefa: Good catch. I did this because I was originally calling it in the post/patch endpoints before I found out all the epafs aren't updated right away. I'll change it back to how it was.

src/db/oracledb/jobs-dao.js
Ruefa: a poorly worded comment. I made it more clear

src/db/oracledb/jobs-dao.js
Ruefa: It is probably a 500 error then. I got the error text from the oracledb error message.

src/db/oracledb/jobs-dao.js
```diff
@@ -56,4 +120,269 @@ const getJobByJobId = async (internalId, jobId) => {
...
+  };
+
+  if (_.includes(additionalFields, 'laborDistribution') && body.laborDistribution !== undefined) {
+    formatLaborDistributionForDb(body, binds);
+  }
+
+  const flattenedBody = flattenBody(body);
+
+  // Some boolean fields need to be massaged to be compatible with db
+  if (_.includes(additionalFields, 'accruesLeaveInd')
```
tsoliangwu0130, line 181: Probably better to follow the same multi-if statements style as we do in Python: https://github.com/osu-mist/style-guides/blob/master/Python.md#multiline-if-statements

src/db/oracledb/jobs-dao.js
```diff
@@ -56,4 +120,269 @@ const getJobByJobId = async (internalId, jobId) => {
...
+ * @param {string} osuId OSU ID of a person
+ * @param {object} body Request body
+ * @returns {string} Query result, null if success
+ */
+const studentCreateJob = async (connection, osuId, body) => {
+  const binds = standardBinds(osuId, body, studentBinds);
+
+  const { outBinds: { result } } = await connection.execute(contrib.studentCreateJob(binds), binds);
+  return result;
+};
```
tsoliangwu0130, line 294: `studentUpdateJob` and `studentCreateJob` can be combined as you did for `graduateJob`.

src/db/oracledb/jobs-dao.js
```diff
@@ -56,4 +120,269 @@ const getJobByJobId = async (internalId, jobId) => {
...
+    contrib.validateChangeReasonCode(),
+    { changeReasonCode },
+  );
+  return rows[0].count > 0;
+};
+
+/**
+ * Determines which Epaf to execute based on fields in body
+ *
+ * @param {string} update 'create' to create a new record, 'update' to update an existing one
```
tsoliangwu0130, line 334: Probably shouldn't call it `update` anymore? maybe `operation` or something else?

src/db/oracledb/jobs-dao.js
```diff
@@ -56,4 +120,269 @@ const getJobByJobId = async (internalId, jobId) => {
...
+        if (changeReasonCode === 'NONE') {
+          result = await updateLaborChangeJob(connection, osuId, body);
+        } else if (changeReasonCode === 'BREAP') {
+          if (employmentType === 'student') {
+            result = await studentUpdateJob(connection, osuId, body);
+          } else if (employmentType === 'graduate') {
+            result = await graduateJob(connection, osuId, body, update);
+          }
+        } else {
+          await updateJob(connection, osuId, body);
```
tsoliangwu0130, line 368: Do we need to return anything?

src/db/oracledb/jobs-dao.js
```diff
@@ -56,4 +120,269 @@ const getJobByJobId = async (internalId, jobId) => {
...
+ * @param {string} update 'create' to create a new record, 'update' to update an existing one
+ * @param {string} osuId OSU ID of a person
+ * @param {object} body Request body
+ * @param {string} internalId Internal ID of a person
+ * @returns {object} raw job record
+ */
+const createOrUpdateJob = async (update, osuId, body) => {
+  const connection = await getConnection('banner');
+  try {
+    let result;
```
tsoliangwu0130, line 343: Sounds like this is an error for the result so I recommend renaming it as `error` or something similar.

src/db/oracledb/jobs-dao.js
```diff
@@ -56,4 +120,269 @@ const getJobByJobId = async (internalId, jobId) => {
...
+      } else if (employmentType === 'student') {
+        result = await studentCreateJob(connection, osuId, body);
+      } else if (employmentType === 'graduate') {
+        result = await graduateJob(connection, osuId, body, update);
+      }
+    }
+
+    // success when result is null
+    if (!result) {
+      // connection.commit();
```
tsoliangwu0130, line 379: Is it a comment or a testing code?

src/db/oracledb/jobs-dao.js
```diff
@@ -56,4 +120,269 @@ const getJobByJobId = async (internalId, jobId) => {
...
+      return result;
```
tsoliangwu0130, line 380: Based on the JSDoc, the returned value should be an object of raw job record, but it looks like we're only returning error or null (if succeed) here?

src/db/oracledb/jobs-dao.js
Ruefa: Testing, shouldn't have been comitted.

src/db/oracledb/jobs-dao.js
Ruefa: Good catch. forgot to update the jsdoc

openapi.yaml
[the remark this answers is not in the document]
Ruefa: It is for both post and patch. I fixed it so the enum doesn't show up on the GET body

src/db/oracledb/jobs-dao.js
```diff
@@ -56,4 +120,257 @@ const getJobByJobId = async (internalId, jobId) => {
...
+      } else if (employmentType === 'graduate') {
+        error = await graduateJob(connection, osuId, body, operation);
+      }
+    }
+
+    if (!error) {
+      connection.commit();
+      return error;
+    }
+    throw new Error(error);
```
tsoliangwu0130, line 370: I guess it should be `return` instead of `throw`?

Ruefa: `return` or `throw` is the difference of returning a `400` or `500` error. I think if we get a db error we should return 500 because the error is likely on our end and not from the users input. Thoughts?

tsoliangwu0130: Make sense.

PR #113, opened 2020-06-08
CO-1870 persons unit tests

Ruefa:


src/tests/unit/persons-dao.test.js
```diff
@@ -0,0 +1,70 @@
...
+  }) => {
+    it(message, () => {
+      const daoProxy = createDaoProxy(daoPath, dbReturn);
+      const result = daoProxy[functionName]({ fakeOsuId });
+      return result.should.eventually.be.fulfilled.and.deep.equal(expected);
+    });
+  });
+
+  it('createPerson should throw an error when outId contains an error', () => {
+    const daoProxy = createDaoProxy(daoPath, { outBinds: { outId: fakeOsuId }, rows: [null] });
```
tsoliangwu0130, line 58: Does `[null]` mean an error? If so, I recommend changing it to `["error"]` or something similar to make it clearer.

src/tests/unit/persons-dao.test.js
```diff
@@ -0,0 +1,70 @@
...
+    const result = daoProxy.createPerson({});
+    return result.should.eventually.be.rejectedWith('Person creation failed');
+  });
+
+  it('createPerson should return error when outId contains ERROR', () => {
+    const daoProxy = createDaoProxy(daoPath, { outBinds: { outId: 'ERROR: something happened' } });
+    const result = daoProxy.createPerson({});
+    return result.should.eventually.be.a('error');
```
tsoliangwu0130, line 66: Should be `an`.

src/tests/unit/persons-dao.test.js
Ruefa: The test description was incorrect. Should be more clear now.

PR #124, opened 2020-07-30
CO-1831 persons integration tests

Ruefa:


src/tests/integration/integration_test.py
```diff
@@ -22,94 +20,250 @@ def setup(cls, config_path, openapi_path):
...
-            {'number': 1, 'size': 501, 'expected_status_code': 400}
-        ]
-        nullable_fields = ['owner']
-        for pagination in testing_paginations:
-            params = {f'page[{k}]': pagination[k] for k in ['number', 'size']}
-            expected_status_code = pagination['expected_status_code']
-            resource = (
-                'PetResource' if expected_status_code == 200
-                else 'ErrorObject'
+    def test_get_person_by_id(self, endpoint='/persons'):
```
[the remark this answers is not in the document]
Ruefa: Good point I only did it this way because the skeleton has it this way. I changed `endpoint` to a local variable in the method

openapi.yaml
```diff
@@ -595,7 +595,8 @@ components:
       description: ID of meal plan
       required: true
       schema:
-        type: string
+        type: number
```
tsoliangwu0130, line 598: Why changing to number? ID must be a string base on JSONAPI: https://jsonapi.org/format/#document-resource-object-identification

src/serializers/meal-plans-serializer.js
```diff
@@ -11,6 +11,19 @@ const planResourceType = planResourceProp.type.enum[0];
 const planResourceAttributes = planResourceProp.attributes.properties;
 const planResourceKeys = _.keys(planResourceAttributes);
 
+/**
+ * Prepares raw data for serializer
+ *
+ * @param {object[]} rawMealPlans raw meal plan data
+ */
```
tsoliangwu0130, line 18: What should be returned?

src/tests/integration/utils.py
```diff
@@ -1,5 +1,6 @@
 """Utility class and functions for integration testing"""
 import argparse
+from deepmerge import always_merger
```
tsoliangwu0130, line 3: Looks like it's a third party package? If so, it should be imported in a separated group.

src/tests/integration/utils.py
```diff
@@ -328,12 +398,35 @@ def check_endpoint(self, endpoint, resource, response_code,
 
         nullable_fields = [] if nullable_fields is None else nullable_fields
         schema = self.get_resource_schema(resource)
-        response = self.make_request(endpoint, response_code,
-                                     params=query_params)
+        response = self.make_request(endpoint,
+                                     response_code,
+                                     params=query_params,
+                                     max_elapsed_seconds=12)
```
tsoliangwu0130, line 404: Is there any request take ~10s?

Ruefa: There is a lot of variance especially when running this many tests. Usually all tests pass with 5 seconds but it would make jenkins yell at us too often.

src/serializers/meal-plans-serializer.js
Ruefa: The function doesn't return anything. Do you think it should?

openapi.yaml
Ruefa: Good catch. meal-plan IDs are stored as integers in oracledb so I added a pattern instead like with `osuId`

src/serializers/meal-plans-serializer.js
tsoliangwu0130: Oh, I didn't notice. Feel free to ignore my comment here.

src/tests/integration/utils.py
tsoliangwu0130: I think 10 secs are too slow IMO, I'd recommend using 5 sec for now and let's see how often the integration test will break then adjust base on the result. Do you know which component takes the longest time to response?

Ruefa: There wasn't any specific endpoint that would cause failures. Seems like one query would just hang for around 10 seconds usually. Sometimes 20+ seconds. I just ran some tests with it at 5 seconds and it worked fine so I'll set it back to 5

PR #125, opened 2020-08-27
CO-1897 add email phone number

Ruefa:


openapi.yaml
```diff
@@ -813,6 +813,16 @@ components:
           nullable: true
           description: ONID username
           example: smithj
+        emailAddress:
+          type: string
+          nullable: true
+          description: Primary email address
+          example: "smithj@example.com"
+        phoneNumber:
```
tsoliangwu0130, line 821: Not sure should we call them primary email/phone or something. Thoughts?

PR #140, opened 2021-03-02
CO-1952 update bhire4 5 workflow

Ruefa:


src/api-routes/persons/{osuId}/jobs.js
```diff
@@ -45,6 +56,10 @@ const post = async (req, res) => {
       return errorBuilder(res, 409, 'A job with the specified job ID already exists.');
     }
 
+    if (changeReasonCode !== 'AAHIR' && !studentEmployeeInd) {
+      return errorBuilder(res, 400, ['AAHIR change reason code must be used to create graduate employee records.']);
+    }
+
```
tsoliangwu0130, line 61: Should we also catch if `AAHIR` is being used to create student job?

src/api-routes/persons/{osuId}/jobs/{jobId}.js
```diff
@@ -48,6 +49,10 @@ const patch = async (req, res) => {
       return errorBuilder(res, 409, 'Job Id in path does not match job Id in body');
     }
 
+    if (changeReasonCode === 'AAHIR' && !studentEmployeeInd) {
+      return errorBuilder(res, 400, ['AAHIR change reason code cannot be used to update graduate job records']);
+    }
+
```
tsoliangwu0130, line 55: Correct me if I am wrong, but I guess `AAHIR` can only be used to create graduate record, so a 400 should be returned in PATCH endpoint here if we see `AAHIR` no matter it is a graduate or a student.

src/db/oracledb/jobs-dao.js
```diff
@@ -302,11 +275,11 @@ const studentJob = async (connection, osuId, body, operation) => {
  * @param {string} operation 'create' to create a new record, 'update' to update an existing one
```
tsoliangwu0130: This should be removed since you've changed the input of this function.

tsoliangwu0130 (changes requested):

src/api-routes/persons/{osuId}/jobs.js
tsoliangwu0130: As discussed, we should reach out to the HR to clarify what `changeReasonCode` should be used for student creation and also validate them here.

tsoliangwu0130 (changes requested):

src/db/oracledb/jobs-dao.js
```diff
@@ -336,7 +308,7 @@ const isValidChangeReasonCode = async (connection, changeReasonCode) => {
  * @param {string} internalId Internal ID of a person
  * @returns {Error} returns error or null if no error occurred
```
tsoliangwu0130, line 306: JSDoc here should be updated since some of arguments have been removed.

src/db/oracledb/jobs-dao.js
```diff
@@ -283,11 +256,11 @@ const updateLaborChangeJob = async (connection, osuId, body) => {
  * @param {string} operation 'create' to create a new record, 'update' to update an existing one
```
tsoliangwu0130: This should be removed as well.

src/db/oracledb/jobs-dao.js
```diff
@@ -336,7 +308,7 @@ const isValidChangeReasonCode = async (connection, changeReasonCode) => {
  * @param {string} internalId Internal ID of a person
  * @returns {Error} returns error or null if no error occurred
  */
-const createOrUpdateJob = async (operation, osuId, body) => {
+const createOrUpdateJob = async (osuId, body) => {
```
tsoliangwu0130: Not sure we should still call this function `createOrUpdate`. We don't have an operation parameter for this function anymore, and there is no explicit business logic here to show whether we are going to create or update jobs here (it is entirely handled at the contrib query by the job change reason codes now). Thoughts?

tsoliangwu0130 (changes requested):

src/db/oracledb/jobs-dao.js
Ruefa: I agree but I had trouble finding a better name. I changed it to `handleJob` thoughts?

src/db/oracledb/jobs-dao.js
```diff
@@ -302,13 +301,12 @@ const isValidChangeReasonCode = async (connection, changeReasonCode) => {
 /**
  * Determines which Epaf to execute based on fields in body
  *
- * @param {string} operation 'create' to create a new record, 'update' to update an existing one
  * @param {string} osuId OSU ID of a person
  * @param {object} body Request body
  * @param {string} internalId Internal ID of a person
```
tsoliangwu0130: This should be removed as well.

tsoliangwu0130 (changes requested):

PR #147, opened 2021-07-20
CO-1975 create medical endpoint

Ruefa:


openapi.yaml
```diff
@@ -1841,6 +1937,55 @@ components:
...
+              type: string
+              nullable: false
+              description: Description of medical code
+        codeDate:
+          type: string
+          format: date
+          nullable: true
+          description: Date associated with this medical code
+          example: '1970-01-01'
+        comment:
```
tsoliangwu0130: I probably won't return this field since I've heard they didn't really maintain these comment fields on other Banner tables, so I guess it'll be the same for this table too.

tsoliangwu0130 (approved):

PR #168, opened 2022-08-26
Make suffix optional in POST and remove from PATCH.

fernamig:
Please review. POST is not working.

### DIFF
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -1016,9 +1016,8 @@ components:
           properties:
             code:
               enum: [NOSU]
-    JobPostAttributes:
+    JobSuffixRequired:
       type: object
-      additionalProperties: false
       properties:
         suffix:
           type: string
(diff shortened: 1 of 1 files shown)


openapi.yaml
```diff
@@ -1534,6 +1546,7 @@ components:
         attributes:
           allOf:
             - $ref: '#/components/schemas/JobAttributes'
+            - $ref: '#/components/schemas/JobSuffixRequired'
```
tsoliangwu0130, line 1549: I'm confused. Isn't suffix optional in POST?

openapi.yaml
```diff
@@ -1016,18 +1016,30 @@ components:
...
         suffix:
           type: string
           nullable: false
           maxLength: 2
           description: >-
             Used to uniquely identify a job record for persons who have multiple
-            jobs that share the same position number. For updating existing
-            jobs, this field is required.
+            jobs that share the same position number. This field is required.
+    JobSuffixOptional:
```
tsoliangwu0130, line 1029: Why do we have two suffix attributes defined and the only difference is `nullable: true/false`? I think what you should do are the following:

1. Make suffix `nullable: true` in post attributes.
2. Separate patch attributes from post attributes (you should still reuse common attributes by references).
3. Remove suffix from patch attributes.
4. Also make sure you modify the corresponding API codes including unit tests to reflect the change. For example, we will no longer need to check whether suffixes are matched between path and body for patch.

tsoliangwu0130 (changes requested):

PR #181, opened 2023-08-10
Making suffix option for jobs POST.

fernamig:


openapi.yaml
```diff
@@ -1567,6 +1589,7 @@ components:
             attributes:
               allOf:
                 - $ref: '#/components/schemas/JobPostAttributes'
+#               - $ref: '#/components/schemas/JobSuffixOptional'
```
tsoliangwu0130, line 1592: You didn't really make this optional since you commented this out.

openapi.yaml
```diff
@@ -1016,18 +1016,39 @@ components:
...
+    JobSuffixOptional:
+      type: object
+      properties:
+        suffix:
+          type: string
+          nullable: true
+          maxLength: 2
+          description: >-
+            Used to uniquely identify a job record for persons who have multiple
+            jobs that share the same position number. This field is optional.
```
tsoliangwu0130, line 1040: Why do you need to create two exact same attributes and the only difference is `nullable: true`? Can you think of a better way to improve it?

tsoliangwu0130 (changes requested):
If I remember it correctly, I think the original request was to make the suffix not required for POST, not make it nullable, right?

PR #182, opened 2023-08-18
Feature/co 2141 jobs unit tests

fernamig:


src/tests/unit/jobs-dao.test.js
```diff
@@ -80,4 +81,137 @@ describe('Test jobs-dao', () => {
...
+      const daoProxy = createDaoProxy(dbReturn, dbReturn);
+
+      const result = daoProxy.handleJob(fakeOsuId, postBody);
+      await result.should.eventually.be.fulfilled.and.deep.equal(expected);
+
+      assertStubCalled(expectedCalledType === jobStubType.graduateJob, stubGraduateJob);
+      assertStubCalled(expectedCalledType === jobStubType.studentJob, stubStudentJob);
+      assertStubCalled(expectedCalledType === jobStubType.terminateJob, stubTerminateJob);
+      assertStubCalled(expectedCalledType === jobStubType.updateLaborChangeJob,
+        stubUpdateLaborChangeJob);
```
tsoliangwu0130: Indentation looks weird here. Let's do this instead:

```js
assertStubCalled(
  expectedCalledType === jobStubType.updateLaborChangeJob, stubUpdateLaborChangeJob
);
```

src/tests/unit/jobs-dao.test.js
```diff
@@ -80,4 +81,137 @@ describe('Test jobs-dao', () => {
...
+      const result = daoProxy.handleJob(fakeOsuId, postBody);
+
+      result.should
+        .eventually.be.fulfilled
+        .and.be.instanceOf(Error)
+        .and.have.property('message', expected);
+    });
+  });
+
+  it('handleJob returns error for graduate with with non-termination code', () => {
```
tsoliangwu0130: Extra "with".

src/tests/unit/jobs-dao.test.js
```diff
@@ -80,4 +81,137 @@ describe('Test jobs-dao', () => {
...
+    expected,
+    postBody,
+    expectedCalledType,
+  }) => {
+    it(message, async () => {
+      const stubGraduateJob = sinon.stub(contrib, 'graduateJob');
+      const stubStudentJob = sinon.stub(contrib, 'studentJob');
+      const stubTerminateJob = sinon.stub(contrib, 'terminateJob');
+      const stubUpdateLaborChangeJob = sinon.stub(contrib, 'updateLaborChangeJob');
+      const daoProxy = createDaoProxy(dbReturn, dbReturn);
```
tsoliangwu0130: Why do we want to pass `dbReturn` twice to `createDaoProxy`?

tsoliangwu0130 (changes requested):

src/tests/unit/jobs-dao.test.js
fernamig: OK

src/tests/unit/jobs-dao.test.js
fernamig: Good catch!

src/tests/unit/jobs-dao.test.js
fernamig: Because handleJobs makes multiple calls to connection.execute and I need to specify the required data to continue the execution of the code. The two get jobs tests use dbReturn for the first call and  "{ rows: [{}] }" for second and beyond calls to connection.execute. The handleJobs tests needs to use dbReturn for all calls.

tsoliangwu0130: If you want to redesign it to be a function with consecutive returns, I think the better way is to make the input become an array and put all your returns in it in order. That way, the function would have more flexibility (could have more than 2 returns, if needed) and is also easier to understand. For this change, I would also add a document string to make it clear.

fernamig: The refactoring for change was minimal.

tsoliangwu0130: What do you mean? I think code quality and readability are very important and NOT minimal.

fernamig: I meant the refactoring of tests to adapt to the change of allowing a variable length of DB results.

src/tests/unit/jobs-dao.test.js
```diff
@@ -16,10 +17,22 @@ describe('Test jobs-dao', () => {
   beforeEach(daoBeforeEach);
   afterEach(() => sinon.restore());
 
-  const createDaoProxy = (dbReturn) => {
+  /**
+   * createDaoProxy creates a proxy DAO using an array of results. dbReturns specifies the result
+   * for the specific call index. The last entry of dbReturns is repeated for calls beyond
+   * what is specifed in dbReturns.
+   *
+   * @param {object} dbReturns an array of results
```
tsoliangwu0130: The input param is an array of objects.

src/tests/unit/jobs-dao.test.js
```diff
@@ -16,10 +17,22 @@ describe('Test jobs-dao', () => {
...
+   * @returns {object} a proxy
```
tsoliangwu0130: A proxy DAO.

src/tests/unit/jobs-dao.test.js
```diff
@@ -16,10 +17,22 @@ describe('Test jobs-dao', () => {
...
+   */
+  const createDaoProxy = (dbReturns) => {
     const executeStub = sinon.stub();
-    executeStub.returns({ rows: [{}] });
-    executeStub.onCall(0).returns(dbReturn);
+
+    executeStub.returns(dbReturns[dbReturns.length - 1]);
+    for (let i = 0; i < dbReturns.length; i += 1) {
```
tsoliangwu0130: Use `_.forEach()` to improve this piece of code.

tsoliangwu0130 (changes requested):

tsoliangwu0130 (approved):

